PR #2, opened 2018-04-23
Add call to strip PNG metadata from a file

SpaceManiac:
This will make it possible to construct an `/icon` full of states, save it out, strip the metadata, and load it back for use as a spritesheet, which will let us avoid sending lots of small images, improving the asset loading experience.

Still nailing down the DM side, but it's looking like:
* 705 pipe images -> 1 spritesheet
* 595 research design icons -> 3 spritesheets
* 116 OOC emoji -> 1 spritesheet
* 27 PDA icons -> 1 spritesheet
* maybe a few others


optimumtact:
is the metadata stripping really needed? They're just pngs with zdata headers, they should load fine in a browser anyway.

SpaceManiac:
`file("some_file.png")` will only give us the first icon_state if the metadata is present.

src/dmi.rs
```diff
@@ -0,0 +1,24 @@
+use std::fs::File;
+use png::{Decoder, Encoder, HasParameters};
+
+use error::Result;
+
+byond_fn! { strip_dmi_metadata(path) {
```
neersighted: I'm trying to prefix functions with their relevant module, so this should be `dmi_strip_metadata`, or similar.

src/error.rs
```diff
@@ -16,6 +18,10 @@ pub enum Error {
     Io(#[cause] io::Error),
     #[fail(display = "Invalid algorithm specified.")]
     InvalidAlgorithm,
+    #[fail(display = "{}", _0)]
```
neersighted: Name these `Image{Decoding,Encoding}`

src/lib.rs
```diff
@@ -12,3 +13,4 @@ pub mod file;
 pub mod hash;
 pub mod log;
 pub mod url;
+pub mod dmi;
```
neersighted: Alphabetical order, please.

src/dmi.rs
```diff
@@ -0,0 +1,24 @@
...
+    strip_metadata(path).err()
+} }
+
+fn strip_metadata(path: &str) -> Result<()> {
+    // Read in the PNG, discarding metadata
+    let (info, mut reader) = Decoder::new(File::open(path)?).read_info()?;
+    let mut buffer = vec![0; reader.output_buffer_size()];
+    reader.next_frame(&mut buffer)?;
+    drop(reader);
```
neersighted: Is there really any reason to explicitly call `mem::drop` here?

SpaceManiac: We need to drop the `File` that was opened for reading so that we can open it for writing below.

PR #5, opened 2018-07-03
For consideration: add a fully-featured html to text rendering function to rust_g

vuonojenmustaturska:
Just to get this out of the way, `html2text` is a huge dependency as it pulls html5ever. It will bloat the resulting binary to 1200 KB-ish. However, some of it is stuff we could use elsewhere maybe (serde), and it does not have any system-side dependencies, being pure rust.

It would give us a much much cleaner way of dealing with untrusted user input, rendering raw input `<span class='narsie'>Borgname</span>` into `Borgname` and leaving `I am a b'anana` as-is. Right now the first one eventually prints out as `<span class=&#39;narsie&#39;>Borgname</span>` and the latter becomes `I am a b&#39;anana`.

As for the branch name, I first intended to use html5ever directly however I'm afraid my rust skills don't stretch that far.


vuonojenmustaturska:
other alternatives: adding simple tag stripping

SpaceManiac:
Getting messages like "I am a b&#39;anana" in the chat has way more to do with double-encoding than with stripping HTML, I think we may need an encoding discipline more than this

PR #6, opened 2018-07-03
Adds .travis.yml

Cyberboss:
Simplest ever


SpaceManiac:
Should be `--release` on both of these since that's what we tell people to do

PR #12, opened 2018-10-15
Making append function

### LINKED ISSUE #8: text2file override is incorrect
This override in the DM API file is incorrect:

```dm
#define text2file(text, fname) rustg_file_write(text, fname)
```

`rustg_file_write` overwrites, while `text2file` appends.

Low priority because /tg/ does not currently use the `file` module or the `RUSTG_OVERRIDE_BUILTINS` feature.

Sindorman:
This PR adds append function for the file. Fixes #8

### DIFF
--- a/src/file.rs
+++ b/src/file.rs
@@ -12,6 +12,10 @@ byond_fn! { file_write(data, path) {
     write(data, path).err()
 } }
 
+byond_fn! { file_append(data, path) {
+    append(data, path).err()
+} }
+
 fn read(path: &str) -> Result<String> {
     let mut file = File::open(path)?;
     let metadata = file.metadata()?;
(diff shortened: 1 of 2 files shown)


SpaceManiac:
I think I would prefer if we added a new function `file_append` and changed the `text2file` override to use that. There's some use in keeping the overwriting version around (namely, it can replace any `fdel`-`text2file` sequence). Thoughts?

Sindorman:
I have got no problem with it. @SpaceManiac . Changed.

SpaceManiac:
Looks good so far. The new function needs an entry in `build.rs` under the `// module: file` section, and the `text2file` override should use that new function instead.

src/file.rs
```diff
@@ -26,3 +31,9 @@ fn write(data: &str, path: &str) -> Result<usize> {
 
     Ok(file.write(data.as_bytes())?)
 }
+
+fn append(data: &str, path: &str) -> Result<usize> {
+    let file = OpenOptions::new().append(true).create(true).open(path)?;
```
SpaceManiac: Needs to be mutable for the `write()` call to work:
```suggestion
    let mut file = OpenOptions::new().append(true).create(true).open(path)?;
```
I'll tweak Travis to catch this.

PR #15, opened 2018-10-23
The function I need to finish out of repo maps except I don't know how to code in rust

Cyberboss:
Yeah, we've got two map PRs open and I don't want to merge either of them. Help getting this working is appreciated

@ninjanomnom @SpaceManiac 

10$ bounty


SpaceManiac:
Repo download seems to work right, haven't tested jobs system yet

Cyberboss:
So progress: It works... occasionally
image
Nothing was changed between these runs (I was just spamming Stop/Go in DD till it stopped panicing)

When I change rust-g to run as a cli app like this it never fails
```rs
fn main() {
    let mut rev = "master";
    let jobId = jobs::start(move || {
        if rev.is_empty() {
            // default
            rev = "FETCH_HEAD".into();
        }
        match git::get_repository_at_reference("test", "git://github.com/typemaker-lang/typemaker", &rev) {
            Ok(()) => "OK".to_owned(),
            Err(e) => e.to_string(),
        }
    });
    loop{
        println!("{}", jobs::check(&jobId));
    }
}
```
So something is being lost in translation at the byond/rust border i'm guessing

Cyberboss:
This isn't necessary with the HTTP stuff, still need a .zip handler

PR #16, opened 2018-12-20
Add Cargo.lock

SpaceManiac:
We've had three problems with dependency shifts in the last week:

* `backtrace v0.3.10` introduced an ambiguity in the `failure` crate, fixed in `v0.3.11`, which caused a spurious Travis failure
* `rustc-demangle v0.1.10` was non-stable-compatible, fixed in `v0.1.11`, which caused a spurious Travis failure
* `backtrace v0.3.13` added a dependency on `autocfg v0.1.1` which bumped minimum rustc version to 1.30.0, tripping up Hippie live which clean-compiled rust-g without updating rustc

We should check in the lockfile to avoid these kinds of issues in the future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 
 /target
 **/*.rs.bk
-Cargo.lock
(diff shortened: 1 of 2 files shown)


Cyberboss:
always :clap: pin :clap: your :clap: dependencies

PR #20, opened 2020-01-04
Optional formatting parameter for log

actioninja:
As title, for replay logging.

This is probably a horrible way of doing this, but I'm not really familiar with Rust. It works though.

### DIFF
--- a/src/log.rs
+++ b/src/log.rs
@@ -24,7 +24,7 @@ byond_fn! { log_write(path, data, format) {
             Entry::Vacant(elem) => elem.insert(open(path)?),
         };
 
-        let format_bool: bool = format.parse().unwrap();
+        if let Some(format_bool) = format.parse() {
         if format_bool {
             // write first line, timestamped
             let mut iter = data.split('\n');
--- a/src/log.rs
+++ b/src/log.rs
@@ -14,7 +14,7 @@ thread_local! {
     static FILE_MAP: RefCell<HashMap<OsString, File>> = RefCell::new(HashMap::new());
 }
 
-byond_fn! { log_write(path, data) {
+byond_fn! { log_write(path, data, format) {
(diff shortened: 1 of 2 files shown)


src/log.rs
```diff
@@ -24,15 +24,20 @@ byond_fn! { log_write(path, data) {
             Entry::Vacant(elem) => elem.insert(open(path)?),
         };
 
-        // write first line, timestamped
-        let mut iter = data.split('\n');
-        if let Some(line) = iter.next() {
-            write!(file, "[{}] {}\n", Utc::now().format("%F %T%.3f"), line)?;
-        }
+        let format_bool: bool = format.parse().unwrap();
```
Cyberboss: I don't know rust so take this with a keg of salt, but isn't it possible for this to throw?

spookydonut:
```suggestion
        if let Some(format_bool) = format.parse() {
```

src/log.rs
```diff
@@ -24,15 +24,22 @@ byond_fn! { log_write(path, data) {
             Entry::Vacant(elem) => elem.insert(open(path)?),
         };
 
-        // write first line, timestamped
-        let mut iter = data.split('\n');
-        if let Some(line) = iter.next() {
-            write!(file, "[{}] {}\n", Utc::now().format("%F %T%.3f"), line)?;
-        }
+        if let Some(format_bool) = format.parse() {
```
spookydonut:
```suggestion
        if let Ok(format_bool) = format.parse() {
```

spookydonut: although actually you should maybe deal with a possible error from the Result

SpaceManiac: suggestion: `if let Ok(false) = format.parse()` to do without formatting, to both avoid the duplicate branch and to make the old behavior the default

actioninja:
hmm yeah ok
image

PR #23, opened 2020-02-15
Ports Skull132's HTTP system

AffectedArc07:
Adds in a HTTP system for use in a PR over at TG to auto-role discord members when they link their accounts. Could probably replace some `world/Export` calls in the future


src/jobs.rs
```diff
@@ -0,0 +1,62 @@
...
+    rx: mpsc::Receiver<Output>,
+    handle: thread::JoinHandle<()>,
+}
+
+type Output = String;
+type JobId = String;
+
+const NO_RESULTS_YET: &str = "NO RESULTS YET";
+const NO_SUCH_JOB: &str = "NO SUCH JOB";
+const JOB_PANICKED: &str = "JOB PANICKED";
```
skull132: You need to either change this to _have_ underscores, or change the defines above to _not_ have underscores. It's an error that still exists in my impl.

AffectedArc07: Ill fix the defines so I dont have to rebuild RUSTG in my other PR

optimumtact:
@SpaceManiac review beg

SpaceManiac:
nothing other than style issues on a code read, will test when able

PR #26, opened 2020-04-10
Add Github Actions building

steamp0rt:
Pre-built binaries for both Linux and Windows.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,0 +1,29 @@
+name: rust-g
+on: [push]
+jobs:
+  build-windows:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions-rs/cargo@v1
+        with:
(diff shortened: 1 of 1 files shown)


steamp0rt:
proof that this works: https://github.com/steamp0rt/rust-g/actions/runs/75508515

.github/workflows/rust.yml
```diff
@@ -0,0 +1,29 @@
...
+  build-windows:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions-rs/cargo@v1
+        with:
+          target: stable-i686-pc-windows-msvc
+          command: build
+          args: --release
+      - uses: actions/upload-artifact@v1
```
SpaceManiac, line 13: What happens for commits that aren't tags? Do the binaries still get built? Where do they go?

steamp0rt: yeah they're still under the actions tab

PR #28, opened 2020-05-04
Adds shim for rust noise library, exposing 2d perlin noise functions

4dplanner:
Provides an interface to get 2d perlin noise generated by the [rust noise library](https://docs.rs/noise/0.6.0/noise/).

 Gives one function: 

- rustg_perlin_noise_get_at_coordinates(seed, x, y): takes an integer seed and floating-point coordinates, and returns a floating-point value in the range [0,1].

benchmarked with this code:
```
/client/New()
	foo()

/proc/foo()
	set waitfor = 0
	world.log << "hello"


	var/start_time = world.timeofday
	var/list/heights = list()
	for(var/x = 0; x < 255; x++)
		for(var/y = 0; y < 255; y++)
			var/height = rustg_perlin_noise_get_at_coordinates("123123", "[x/255]", "[y/255]")
			if(x == 239)
				world.log << height
			heights.Add(height)

	var/end_time = world.timeofday
	world.log << "we took [end_time - start_time] to generate [length(heights)] values"
```

giving on my pc a result of `we took 4 to generate 65025 values`


SpaceManiac:
Any objections to making it one function that accepts (seed, x, y)? The `Perlin` struct can continue to be cached, but I don't see a need to introduce formal state

SpaceManiac:
Functionality looks good, will ponder whether I care about the naming overnight

PR #31, opened 2020-06-02
Make code more idiomatic to standard Rust

Jared-Fogle:
I do Rust as part of my living, and I wanted to update this codebase to match with the standard way Rust repositories are managed. This PR is mostly running Rustfmt and fixing Clippy errors. These are both invaluable tools I highly recommend you try out if you haven't already. Rustfmt is an autoformatter that not only makes it so there's no conflicting styles, but also makes it really easy to do big refactors without having to care about style too much. VS Code has an option to run a formatter when you save, and if you're using a Rust LSP (RLS or the newer rust-analyzer), then it helps productivity a lot. Clippy is a very good linter that finds issues that the normal rustc doesn't catch. For this repo, it found some style issues and a minor performance issue that were easily fixed up.

A lot of the code is inside macros, which clippy and Rustfmt can't introspect very well, but this covers most of it.

- Runs cargo fmt*
- Fixes error brought up by cargo clippy*
- Upgrades to Rust 2018, which is regularly updated
- Change `const _: () = "bla"` to `compile_error!("bla")`
- Change `use` to be use the idiomatic grouped syntax

\* - I highly recommend setting up a GitHub action to do these.

### DIFF
--- a/src/byond.rs
+++ b/src/byond.rs
@@ -10,11 +10,11 @@ thread_local! {
     static RETURN_STRING: RefCell<CString> = RefCell::new(CString::default());
 }
 
-pub fn parse_args<'a>(argc: c_int, argv: &'a *const *const c_char) -> Vec<Cow<'a, str>> {
+pub fn parse_args<'a>(argc: c_int, argv: *const *const c_char) -> Vec<Cow<'a, str>> {
     unsafe {
         slice::from_raw_parts(*argv, argc as usize)
             .iter()
-            .map(|ptr| CStr::from_ptr(*ptr))
+            .map(|ptr| CStr::from_ptr(ptr))
             .map(|cstr| cstr.to_string_lossy())
             .collect()
     }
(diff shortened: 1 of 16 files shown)


Jared-Fogle:
I will take a look at why that made Travis fail in a bit.

SpaceManiac:
> I will take a look at why that made Travis fail in a bit.

Features are not set at compile-time for build scripts. They are passed in environment variables when the build script is run.

SpaceManiac:
Not generally a fan of mass machine formatting but I already spotted two bugs reviewing this so w/e

Jared-Fogle:
Reverted the `enabled!` check in that case.

src/byond.rs
```diff
@@ -1,20 +1,21 @@
...
     static RETURN_STRING: RefCell<CString> = RefCell::new(CString::default());
 }
 
-pub fn parse_args<'a>(argc: c_int, argv: &'a *const *const c_char) -> Vec<Cow<'a, str>> {
+pub fn parse_args<'a>(argc: c_int, argv: *const *const c_char) -> Vec<Cow<'a, str>> {
     unsafe {
         slice::from_raw_parts(*argv, argc as usize)
             .iter()
-            .map(|ptr| CStr::from_ptr(*ptr))
+            .map(|ptr| CStr::from_ptr(ptr))
```
SpaceManiac, line 18: My fault as well for not catching it the first time but I'd like to point out now that this change was wrong. See 6ba75cd

Jared-Fogle: Weird, wonder what the actual solution to fixing this lint is then. I did build this for myself to try out and I didn't notice any issues, so maybe I just didn't test enough.

PR #35, opened 2020-07-25
Updates file reading to match BYOND

Crossedfall:
BYOND strips `\r` characters when converting a file's contents to a string. Doing so here will maintain feature parity and makes the override sensible.


Crossedfall:
Double checked and confirmed that appears to be the case. Updated accordingly.

SpaceManiac:
It might make sense to add an option to get the "preserve CR" behavior back in the future

PR #38, opened 2020-09-19
Various updates and improvements!

aspenluxxxy:
* Update all crates to latest versions.  
* Use a `DashMap` over a `RwLock<HashMap>` for SQL pooling. 
  * Also, use a `usize` as the key value instead of a string of the `usize`
* Use `once_cell` over `lazy_static` - no macros needed. 
  * The `AtomicUsize` used for pool IDs is no longer wrapped in a lazy_static, it's just it's own thing now.
* Add `codegen-units = 1` to the release profile. This allows the LLVM to optimize the code more, at the cost of multi-threaded compile time.

I've done a quick test, SQL appears to work correctly.

### DIFF
--- a/src/http.rs
+++ b/src/http.rs
@@ -1,5 +1,7 @@
 use crate::{error::Result, jobs};
+use once_cell::sync::Lazy;
 use std::collections::{BTreeMap, HashMap};
+use std::io::Write;
 
 // ----------------------------------------------------------------------------
 // Interface
@@ -58,10 +60,10 @@ byond_fn! { http_check_request(id) {
 const VERSION: &str = env!("CARGO_PKG_VERSION");
 const PKG_NAME: &str = env!("CARGO_PKG_NAME");
 
-fn setup_http_client() -> reqwest::Client {
+fn setup_http_client() -> reqwest::blocking::Client {
     use reqwest::{
+        blocking::Client,
         header::{HeaderMap, USER_AGENT},
-        Client,
     };
 
     let mut headers = HeaderMap::new();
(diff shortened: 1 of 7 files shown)


src/http.rs
```diff
@@ -137,18 +137,21 @@ fn submit_request(prep: RequestPrep) -> Result<String> {
...
+    let headers = response.headers().clone();
+    for (key, value) in headers.iter() {
+        if let Ok(value) = value.to_str() {
+            resp.headers.insert(key.as_str(), value);
+        }
+    }
+
     if let Some(output_filename) = prep.output_filename {
-        std::io::copy(&mut response, &mut std::fs::File::create(&output_filename)?)?;
+        let mut writer = std::io::BufWriter::new(std::fs::File::create(&output_filename)?);
```
aspenluxxxy, line 148: `BufWriter`s are usually _much_ better in this situation, unbuffered I/O has a performance penalty.

src/lib.rs
```diff
@@ -20,13 +20,10 @@ extern crate reqwest;
 #[cfg(any(feature = "http", feature = "sql"))]
```
aspenluxxxy, line 20: Dunno why we still have `extern crate` definitions, we're using Rust 2018 now.

src/sql.rs
```diff
@@ -100,10 +97,8 @@ byond_fn! { sql_check_query(id) {
 // ----------------------------------------------------------------------------
 // Main connect and query implementation
 
-lazy_static! {
-    static ref POOL: RwLock<HashMap<String, Pool>> = Default::default();
-    static ref NEXT_ID: std::sync::atomic::AtomicUsize = Default::default();
-}
+static POOL: Lazy<DashMap<String, Pool>> = Lazy::new(DashMap::new);
+static NEXT_ID: AtomicUsize = AtomicUsize::new(0);
```
aspenluxxxy, line 101: No clue why an AtomicUsize was in a lazy_static. That was redundant and probably had a tiny performance penalty.

SpaceManiac (approved):
seems fine

optimumtact:
are you waiting for someone else to +1?

SpaceManiac:
> are you waiting for someone else to +1?

no, I'm just slow

PR #39, opened 2020-10-04
Add dmi_resize_png using the image crate

ShadowLarkens:
Please be gentle this is literally the first time I've ever touched rust ;w;
I am probably going to be absolutely helpless if this needs any optimization or tweaks to file writing

### DIFF
--- a/src/dmi.rs
+++ b/src/dmi.rs
@@ -10,7 +10,19 @@ byond_fn! { dmi_strip_metadata(path) {
 } }
 
 byond_fn! { dmi_create_png(path, width, height, data) {
-    create_png(path, width, height , data).err()
+    create_png(path, width, height, data).err()
+} }
+
+byond_fn! { dmi_resize_png(path, width, height, resizetype) {
+    let resizetype = match resizetype {
+        "catmull" => image::imageops::CatmullRom,
+        "gaussian" => image::imageops::Gaussian,
+        "lanczos3" => image::imageops::Lanczos3,
+        "nearest" => image::imageops::Nearest,
+        "triangle" => image::imageops::Triangle,
+        _ => image::imageops::Nearest,
+    };
+    resize_png(path, width, height, resizetype).err()
 } }
 
 fn strip_metadata(path: &str) -> Result<()> {
@@ -62,4 +74,15 @@ fn create_png(path: &str, width: &str, height: &str, data: &str) -> Result<()> {
     encoder.set_depth(png::BitDepth::Eight);
     let mut writer = encoder.write_header()?;
     Ok(writer.write_image_data(&result)?)
+}
+
+fn resize_png<P: AsRef<Path>>(path: P, width: &str, height: &str, resizetype: image::imageops::FilterType) -> std::result::Result<(), Error> {
+    let width = u32::from_str_radix(width, 10)?;
+    let height = u32::from_str_radix(height, 10)?;
+
+    let img = image::open(path.as_ref())?;
+
+    let newimg = img.resize(width, height, resizetype);
+
+    Ok(newimg.save_with_format(path.as_ref(), image::ImageFormat::Png)?)
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -115,6 +115,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e3b5ca7a04898ad4bcd41c90c5285445ff5b791899bb1b0abdd2a2aa791211d7"
 
 [[package]]
+name = "bytemuck"
+version = "1.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "41aa2ec95ca3b5c54cf73c91acf06d24f4495d5f1b1c12506ae3483d646177ac"
+
+[[package]]
 name = "byteorder"
 version = "1.3.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -583,6 +589,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "gif"
+version = "0.11.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "02efba560f227847cb41463a7395c514d127d4f74fff12ef0137fff1b84b96c4"
+dependencies = [
+ "color_quant",
+ "weezl",
+]
+
+[[package]]
 name = "git2"
 version = "0.13.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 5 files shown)


optimumtact:
Thanks for the PR, and welcome aboard 👍

src/dmi.rs
```diff
@@ -63,3 +67,13 @@ fn create_png(path: &str, width: &str, height: &str, data: &str) -> Result<()> {
     let mut writer = encoder.write_header()?;
     Ok(writer.write_image_data(&result)?)
 }
+
+fn resize_png(path: &str, width: &str, height: &str) -> Result<()> {
+    let width = u32::from_str_radix(width, 10)?;
+    let height = u32::from_str_radix(height, 10)?;
+
+    let img = image::open(path).unwrap();
```
SpaceManiac, line 75: I would much rather return an error using `?` rather than panicking if loading or saving fails

SpaceManiac:
Have you done any tests of how fast this is compared to http://www.byond.com/docs/ref/info.html#/icon/proc/Scale ?

AnturK:
What's the use case here ?

ShadowLarkens:
@AnturK @SpaceManiac The point of this is that Scale() has this little caveat: `Scale() automatically performs antialiasing to avoid unwanted artifacts.`  I have seen examples of this antialiasing *not* occuring, but... I have no idea why. There doesn't appear to be any consistent way to convince BYOND "Hey, don't just randomly apply bilinear antialiasing to this pixel art that I'm trying to upscale."

src/dmi.rs
ShadowLarkens: Doooes that just mean add a question mark to the end of this line? Could you make a code suggestion? I'm really unfamiliar with rust's syntax and error handling.

ShadowLarkens:
Like, if anyone knows what the rhyme or reason to Scale() occasionally just deciding to apply bilinear antialiasing and how to prevent it from doing that, I'd be happy to implement my icon upscaling with native BYOND procs. This idea came up while working with the spritesheet asset datum and needing to upscale a spritesheet for a UI.

ZeWaka:
> Like, if anyone knows what the rhyme or reason to Scale() occasionally just deciding to apply bilinear antialiasing and how to prevent it from doing that, I'd be happy to implement my icon upscaling with native BYOND procs.

You probably don't have `appearance_flags |= PIXEL_SCALE` set if you're performing this operation on an atom. However, that var doesn't seem to exist for /icon or /image, which could be your usecase.

ShadowLarkens:
Yeah, it's direct /icon modification that matters.

src/dmi.rs
ShadowLarkens, line 75:
```suggestion
    let img = image::open(path)?;
```

src/dmi.rs
```diff
@@ -63,3 +67,13 @@ fn create_png(path: &str, width: &str, height: &str, data: &str) -> Result<()> {
...
+    let newimg = img.resize(width, height, image::imageops::Nearest);
+
+    Ok(newimg.save_with_format(&path, image::ImageFormat::Png).unwrap())
```
ShadowLarkens, line 78:
```suggestion
    Ok(newimg.save_with_format(&path, image::ImageFormat::Png)?)
```

ShadowLarkens:
Hrm. Can someone help me with this? Even after reading through the rust error handling guidelines, I can't figure out how to implement non-panick file read/write, as the image library returns [ImageError](https://docs.rs/image/0.23.10/image/error/enum.ImageError.html) instead of Error and there's no type conversion defined for "ImageError -> Error".

ShadowLarkens:
Well it was an adventure, but with PJB's help ~~definitely just doing it for me with how bad I am at rust~~ it finally returns an error and doesn't blow up rust-g when it fails

spookydonut:
@ShadowLarkens new conflicts

ShadowLarkens:
@spookydonut Resolved, let me know if there's anything I need to change to follow #42

src/dmi.rs
```diff
@@ -63,3 +68,32 @@ fn create_png(path: &str, width: &str, height: &str, data: &str) -> Result<()> {
     let mut writer = encoder.write_header()?;
     Ok(writer.write_image_data(&result)?)
 }
+
+#[derive(Debug)]
+enum ResizePngError {
```
spookydonut, line 73: see `pub enum Error`

src/dmi.rs
```diff
@@ -63,3 +68,32 @@ fn create_png(path: &str, width: &str, height: &str, data: &str) -> Result<()> {
...
+    }
+}
+
+impl From<ImageError> for ResizePngError {
+    fn from(error: ImageError) -> Self {
+        ResizePngError::Image(error)
+    }
+}
+
+fn resize_png(path: &str, width: &str, height: &str) -> std::result::Result<(), ResizePngError> {
```
spookydonut, line 90:
```suggestion
fn resize_png(path: &str, width: &str, height: &str) -> std::result::Result<(), Error> {
```

spookydonut (changes requested):

spookydonut:
@aspenluxxxy

src/dmi.rs
```diff
@@ -10,7 +11,11 @@ byond_fn! { dmi_strip_metadata(path) {
 } }
 
 byond_fn! { dmi_create_png(path, width, height, data) {
-    create_png(path, width, height , data).err()
+    create_png(path, width, height, data).err()
+} }
+
+byond_fn! { dmi_resize_png(path, width, height) {
+    resize_png(path, width, height).err().map(|e| format!("{:?}", e))
```
aspenluxxxy, line 18: No, using `rust_g::Error` will handle this

src/dmi.rs
aspenluxxxy, line 90:
```suggestion
fn resize_png<P: AsRef<Path>>(path: P, width: u32, height: u32) -> std::result::Result<(), Error> {
```

src/dmi.rs
```diff
@@ -63,3 +68,32 @@ fn create_png(path: &str, width: &str, height: &str, data: &str) -> Result<()> {
...
+    }
+}
+
+fn resize_png(path: &str, width: &str, height: &str) -> std::result::Result<(), ResizePngError> {
+    let width = u32::from_str_radix(width, 10)?;
+    let height = u32::from_str_radix(height, 10)?;
+
+    let img = image::open(path)?;
+
+    let newimg = img.resize(width, height, image::imageops::Nearest);
```
aspenluxxxy, line 96: Wouldn't it be better to allow choice of resizing method? Nearest Neighbor isn't always the best.

src/dmi.rs
```diff
@@ -63,3 +68,32 @@ fn create_png(path: &str, width: &str, height: &str, data: &str) -> Result<()> {
...
+
+    Ok(newimg.save_with_format(&path, image::ImageFormat::Png)?)
```
aspenluxxxy, line 98:
```suggestion
    Ok(newimg.save_with_format(path.as_ref(), image::ImageFormat::Png)?)
```

src/dmi.rs
aspenluxxxy, line 73: Just add a variant to `rust_g::Error`

aspenluxxxy (changes requested):

src/dmi.rs
ShadowLarkens: Can BYOND call functions with u32 arguments? I thought it explicitly only worked with strings

aspenluxxxy: Correct, but you'd convert them to u32 in the `byond_fn!`

ShadowLarkens: Trying to do that results in the `?` no longer working for the ParseIntErrors because byond_fn! doesn't actually return a Result

ShadowLarkens: That's probably not exactly why it doesn't work, but I have no idea how to make it stop yelling at me with this

Cyberboss (changes requested):
Please fix the merge conflict

ShadowLarkens:
@Cyberboss Done

Cyberboss (approved):

PR #40, opened 2020-10-04
Refactor logging to be based on an asynchronous tokio buffer

aspenluxxxy:


optimumtact:
Isn't this overkill for logging? this will sureyl have a higher memory overhead and rust-g counts towards the memory limits

actioninja:
Putnam was finding that rust-g was causing server stuttering when shitloads of logs got thrown at it. They made a threaded solution in extools.

optimumtact:
I find that very hard to believe

Cyberboss:
This is a good PR, but yeah I'd like to see a benchmark of the slowdown in the old logging method just so we're sure this is the cause

PR #42, opened 2020-10-11
Modernization

aspenluxxxy:
* Use `thiserror` over `failure` - cleaner code.
* Use `flume` instead of `std::mpsc` - better performance, 100% safe
* Use the rust-crypto crates (`md-5`, `sha-1`, `sha2`) over `crypto-hash` - slightly more complex code, but one less openssl dependency (sadly, `mysql` still pulls in `openssl`)
* Get rid of all the `extern crate`s 
* Use `serde` with the "derive" feature over `serde_derive`
* Update `mysql` to 20.0
* Make the `file_read`/`append`/`write` used buffered IO, `write`/`append` will ensure that filesystem metadata is synchronized.


spookydonut:
@aspenluxxxy conflicts

aspenluxxxy:
@spookydonut fixed

PR #43, opened 2020-10-11
Adds file_exists function

### LINKED ISSUE #41: fexists() implementation
`fexists("path")` on Linux return 0 for symlinked files.

_why_

Who knows but it'd be useful for some /tg/ prod shared config stuff.

Can we get a `rustg_fexists(path)` call please?

ZeWaka:
Implements a function to check if a file exists or not.
Returns `"true"` or `"false"`.

Tested and works, will follow symlinks (on Windows, at least).

Closes #41

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ The default features are:
 Additional features are:
 * noise: 2d Perlin noise.
 * url: Faster replacements for `url_encode` and `url_decode`.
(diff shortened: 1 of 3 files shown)


Cyberboss:
Based PR. @MrStonedOne worth adding because of that symlink/fexists() nonsense?

PR #47, opened 2020-10-12
Update README to use --target instead of toolchain overrides

PJB3005:
Toolchain overrides have no reason and just create a confusing mess for stuff like this.


SpaceManiac:
Thanks, a couple people had run into this recently & I hadn't got around to updating the readme yet

PR #49, opened 2020-10-14
Add more to GitHub workflow, make it support PRs

Jared-Fogle:
Ideally replaces Travis.

Closes #48


Jared-Fogle:
oh god its the error i dont know how to solve

SpaceManiac:
use `ldd` to find out which dependency is missing and install it as part of the script

Jared-Fogle:
image
image
Ready. @spookydonut was saying something about CI times over 10 minutes, and I believe those also don't run Windows and Linux side by side.

spookydonut:
> 
> 
> image
> image
> Ready. @spookydonut was saying something about CI times over 10 minutes, and I believe those also don't run Windows and Linux side by side.

uncached travis jobs we something like 10 minutes for windows and 14 minutes for linux

Cyberboss (approved):

.github/workflows/rust.yml
```diff
@@ -28,11 +40,33 @@ jobs:
           sudo dpkg --add-architecture i386
           sudo apt-get update
           sudo apt-get install g++-multilib zlib1g-dev:i386 libssl-dev:i386 pkg-config:i386
+          env BYOND_MAJOR=513 BYOND_MINOR=1521 ./scripts/install_byond.sh
```
ZeWaka, line 43: not latest?

Jared-Fogle: I copied this from the old Travis script, not sure if there's an immediately better option.

ZeWaka: I mean, the TG dockerfile uses [1533](https://github.com/tgstation/tgstation/blob/master/Dockerfile) at least. 1536 would also work I presume (latest at time of comment, what goon runs).

Cyberboss: Either way, i'd set these higher up in a global env section

Cyberboss (changes requested):
Put BYOND version in global `env:` section

.github/workflows/rust.yml
```diff
@@ -33,14 +33,16 @@ jobs:
   build-linux:
     runs-on: ubuntu-latest
     env:
+      BYOND_MAJOR: 513
+      BYOND_MINOR: 1521
```
ZeWaka, line 37: still 🤔

Cyberboss: vOv

Cyberboss (approved):

.github/workflows/rust.yml
ZeWaka: literally using that version for no reason, even the /tg/ dockerfile is newer...

Cyberboss: https://github.com/tgstation/rust-g/issues/56

PR #51, opened 2020-10-23
Spice up the buildscript by splitting the module dm components into their own files

vuonojenmustaturska:
Seems cleaner and more intuitive to me this way, also I kind of disliked the "hardcoded" dm contents in strings.


Cyberboss:
based

aspenluxxxy (approved):
lgtm

PR #54, opened 2020-10-26
Adds a jobs feature to make rust_g.dm work again

vuonojenmustaturska:
In #51 I foolishly assumed `jobs` was a feature in line with everything else, turns out it wasn't and now the jobs defines are missing from rust_g.dm.

~~I wonder if feature-interdependencies are possible...~~

They are. `jobs` is now a feature depended on by `http` and `sql`.


Cyberboss (approved):

vuonojenmustaturska:
oh yeah this is tested to compile with default features and `--features="dmi git log json noise" --no-default-features`, `jobs` is not directly a default feature but a dependency of default features `http` and `sql`

Cyberboss (approved):

Cargo.toml
```diff
@@ -17,7 +17,7 @@ lto = true
 
 [dependencies]
 thiserror = "1.0"
-flume = "0.9"
+flume = {version = "0.9", optional = true }
```
Cyberboss, line 20:
```suggestion
flume = { version = "0.9", optional = true }
```
_eye twitch_

Cyberboss (changes requested):

Cyberboss (approved):

SpaceManiac:
hmm today I will click "Rebase Merge"

Cyberboss:
Why is it the default then?

SpaceManiac:
it remembers what you used last

PR #57, opened 2020-11-10
adds new type of noise based on cellular automata.

EdgeLordExe:
Adds new type of noise

to generate the noise you first must generate the map which you call **rustg_cnoise_generate(precentage, smoothing_iterations)**

precentage represents the chance the square is 0 or 1, lower numbers mean more 1's, higher numbers mean more 0's

smoothing_iterations represent how many iterations the algorithm is allowed to do.

returns the grid.


actioninja:
E

Cyberboss:
What's the format of the return value here?

EdgeLordExe:
> What's the format of the return value here?

a string which is filled with 1's and 0's, to access a specific 'cell'look at floyd's pr he does it there.

src/cellularnoise.rs
```diff
@@ -0,0 +1,100 @@
...
+
+byond_fn! { cnoise_generate(precentage,smoothing_iterations, birth_limit, death_limit) {
+    noise_gen(precentage, smoothing_iterations, birth_limit, death_limit).ok()
+} }
+
+fn noise_gen(prec_as_str : &str, smoothing_level_as_str : &str, birth_limit_as_str : &str, death_limit_as_str : &str)-> Result<String> {
+    let prec = prec_as_str.parse::<i32>().expect("parse failed");
+    let smoothing_level = smoothing_level_as_str.parse::<i32>().expect("parse failed");
+    let birth_limit = birth_limit_as_str.parse::<i32>().expect("parse failed");
+    let death_limit = death_limit_as_str.parse::<i32>().expect("parse failed");
```
SpaceManiac, line 12: These 4 should be using `?` instead of `.expect` so they don't crash the whole server on failure

SpaceManiac (changes requested):

src/cellularnoise.rs
```diff
@@ -0,0 +1,100 @@
...
+    //Noise generation
+
+    let mut zplane = vec![vec![0; 254]; 254]; // 254 but we start at 0, and since byond starts at one it is 255 byond wise.
```
SpaceManiac, line 15: - Should be `255`, not `254`
- Shouldn't be hardcoded to 255
- Can use `vec![0u8; ...]` to reduce memory footprint from default of `i32`, or even `vec![false; ...]` for additional clarity

src/cellularnoise.rs
```diff
@@ -0,0 +1,100 @@
...
+    for i in 0..zplane.len() {
+        for j in 0..zplane.len(){
```
SpaceManiac, line 17: Consider using `zplane[i].len()` to support non-square regions

src/cellularnoise.rs
```diff
@@ -0,0 +1,100 @@
...
+            }
+        }
+    }
+
+
+    let mut string:String = String::from("");
+    for i in 0..zplane.len() {
+        
+        for j in 0..zplane.len(){
+           if zplane[j][i] == 1 {
```
SpaceManiac, line 90: Is `[j][i]` intentional here?

src/cellularnoise.rs
```diff
@@ -0,0 +1,100 @@
...
+                string = [string, String::from("1")].join("");
```
SpaceManiac, line 91: Should probably `write!()` to a buffer here rather than use `.join()` on loop

SpaceManiac (changes requested):

src/cellularnoise.rs
Qustinnus: it should be 254 because our maps are 255x255

Qustinnus: agree with the rest though

optimumtact:
https://www.oranges.net.nz/~oranges/cellularautomata.mp4

src/cellularnoise.rs
EdgeLordExe: yeah it iterates through the entire array, the order of the variables doesnt matter.

src/cellularnoise.rs
```diff
@@ -28,53 +28,53 @@ fn noise_gen(prec_as_str : &str, smoothing_level_as_str : &str, birth_limit_as_s
...
+                        zplane[i][j] = false;
                     } else{
-                        zplane[j][i] = true;
+                        zplane[i][j] = true;
                     }
                 }
                 else{
                     if sum > birth_limit{
-                        zplane[j][i] = false;
+                        zplane[i][j] = false;
```
SpaceManiac, line 75: I think you have false/true backwards on this one

Qustinnus: yep he does

EdgeLordExe: that might be the case

Qustinnus:
my body is ready

Qustinnus:
my tg pr is going stale send the marines :(

dmsrc/cellularnoise.dm
```diff
@@ -0,0 +1 @@
+#define rustg_cnoise_generate(precentage, smoothing_iterations, birth_limit, death_limit) call(RUST_G, "cnoise_generate")(precentage, smoothing_iterations, birth_limit, death_limit)
```
Cyberboss, line 1: DMDoc this "proc" please

Cyberboss (changes requested):

dmsrc/cellularnoise.dm
Qustinnus, line 1:
```suggestion
///This proc generates a cellular automata noise grid, returns a string with 0 for open and 1 for closed tiles.
#define rustg_cnoise_generate(precentage, smoothing_iterations, birth_limit, death_limit) call(RUST_G, "cnoise_generate")(precentage, smoothing_iterations, birth_limit, death_limit)
```

Qustinnus, line 1:
```suggestion
///This proc generates a cellular automata noise grid, returns a string with 0 for open and 1 for closed tiles.
#define rustg_cnoise_generate(precentage, smoothing_iterations, birth_limit, death_limit) call(RUST_G, "cnoise_generate")(precentage, smoothing_iterations, birth_limit, death_limit)
```

Cyberboss:
_Full_ dmdoc's please @Qustinnus (Each arg and a detailed desc of the return format)

Qustinnus:
fair one sec

Qustinnus:
@Cyberboss better?

Cyberboss:
`* vars:` isn't necessary but yes

Qustinnus:
yeah dunno looked up docs in code as example

dmsrc/cellularnoise.dm
```diff
@@ -0,0 +1,8 @@
+/* This proc generates a cellular automata noise grid which can be used in procedural generation methods
+* * percentage (required) The chance of a turf starting closed
+* * smoothing_iterations(required) The amount of iterations the cellular automata simulates before returning the results
+* * birth_limit (required) If the number of neighbouring cells is higher than this amount, a cell is born
+* * death_limit (required) If the number of neighbouring cells is lower than this amount, a cell dies
+* * return value: returns a single string that goes row by row, with values of 1 representing an alive cell, and a value of 0 representing a dead cell.
```
Cyberboss, line 6:
```suggestion
/**
 * This proc generates a cellular automata noise grid which can be used in procedural generation methods. Returns a single string that goes row by row, with values of 1 representing an alive cell, and a value of 0 representing a dead cell.
* * percentage (required) The chance of a turf starting closed
* * smoothing_iterations(required) The amount of iterations the cellular automata simulates before returning the results
* * birth_limit (required) If the number of neighboring cells is higher than this amount, a cell is born
* * death_limit (required) If the number of neighboring cells is lower than this amount, a cell dies
```

Cyberboss (changes requested):

PR #64, opened 2021-03-18
Cargo update

ZeWaka:
updates cargo dependency versions to latest and makes code compliant with the changes


src/cellularnoise.rs
```diff
@@ -25,7 +25,7 @@ fn noise_gen(
     let mut zplane = vec![vec![false; width]; height];
     for row in zplane.iter_mut() {
         for cell in row.iter_mut() {
-            *cell = rand::thread_rng().gen_range(0, 100) < percentage;
+            *cell = rand::thread_rng().gen_range(0..=100) < percentage;
```
willox, line 28: rand 0.7.0's documentation states that its range is exclusive of the high parameter. Should this instead be the following?
```suggestion
            *cell = rand::thread_rng().gen_range(0..100) < percentage;
```

ZeWaka: oh weird, it's inclusive of low and exclusive of high, so yes

PR #68, opened 2021-05-03
Adds DM proc to get the rustg version

ZeWaka:
Title, adapted from https://github.com/ParadiseSS13/rust-g/pull/6


Cyberboss:
wtf no feature flag?

ZeWaka:
> wtf no feature flag?

I feel like it's pretty core functionality to get the version of the program you're running, and shouldn't be locked behind an arbitrary feature flag.

PR #71, opened 2021-07-12
Make the url feature default

Mothblocks:
We're using this on tg right now with https://github.com/tgstation/tgstation/pull/60190, and it appears to be much faster and has not yet caused issues.

SpaceManiac has brought it up, and I've agreed for a long time, that having all these individual features was a mistake--I'd eventually like to get rid of them once they're all on by default.


ZeWaka:
Let's see if lummoxes improvement beats this

PR #73, opened 2021-08-11
Stringify url_encode and decode input before sending it to Rust

Gamer025:
Try to stringify all input before sending it to Rust since interop only works with strings
Should be good enough to only stringify inside the rustg_url_encode/decode define since the other defines just refer to that
PR on request of @Mothblocks from https://github.com/tgstation/tgstation/pull/60802

### DIFF
--- a/dmsrc/url.dm
+++ b/dmsrc/url.dm
@@ -1,5 +1,5 @@
-#define rustg_url_encode(text) call(RUST_G, "url_encode")(text)
-#define rustg_url_decode(text) call(RUST_G, "url_decode")(text)
+#define rustg_url_encode(text) call(RUST_G, "url_encode")("[text]")
+#define rustg_url_decode(text) call(RUST_G, "url_decode")("[text]")
 
 #ifdef RUSTG_OVERRIDE_BUILTINS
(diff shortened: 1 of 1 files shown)


Cyberboss (approved):

dmsrc/url.dm
```diff
@@ -1,5 +1,5 @@
-#define rustg_url_encode(text) call(RUST_G, "url_encode")(text)
-#define rustg_url_decode(text) call(RUST_G, "url_decode")(text)
+#define rustg_url_encode(text) call(RUST_G, "url_encode")("[text]")
+#define rustg_url_decode(text) call(RUST_G, "url_decode")("[text]")
```
Cyberboss, line 2:
```suggestion
#define rustg_url_decode(text) call(RUST_G, "url_decode")(text)

```

I'd rather have it error here

Cyberboss (changes requested):

Cyberboss (approved):

Cyberboss:
@Mothblocks merge this before you cut the patch, skip the 24h if you want

Mothblocks:
i always do 😎

PR #74, opened 2021-08-31
Add worley noise optional feature

EdgeLordExe:
Doing worley noise in DM is a performance suicide rust-g is the only option.

# Worley Noise!
This noise works by creating regions with nodes that are later mapped onto a set grid.

Each cell on the grid finds the nearest node in  neighbouring and it's own region, it then finds the second closest and subtracts the distance one from another.

If  it passes a threshold then congratulations, it is set to true.

# Samples
obraz
obraz
obraz
obraz

### DIFF
--- a/src/worleynoise.rs
+++ b/src/worleynoise.rs
@@ -1,7 +1,7 @@
 use crate::error::Result;
 use rand::*;
 use std::fmt::Write;
-use std::thread;
+use std::rc::Rc;
 
 byond_fn! { worley_generate(region_size, threshold, width, height) {
     worley_noise(region_size, threshold, width, height).ok()
@@ -27,9 +27,9 @@ fn worley_noise(str_reg_size : &str, str_positive_threshold: &str, str_width: &s
     for row in map.cell_map{
         for cell in row{
             if cell.value {
-                output.append_str("1");
+                let _ = write!(output, "1");
             } else {
-                output.append_str("0");
+                let _ = write!(output, "0");
             }
         }
     }
--- a/src/worleynoise.rs
+++ b/src/worleynoise.rs
@@ -1,81 +1,204 @@
 use crate::error::Result;
 use rand::*;
 use std::fmt::Write;
+use std::thread;
 
-byond_fn! { worley_generate(density, threshold, width, height) {
-    worley_noise(density, threshold, width, height).ok()
+byond_fn! { worley_generate(region_size, threshold, width, height) {
+    worley_noise(region_size, threshold, width, height).ok()
 } }
 
-//In the worst possible situation this is O(n^3) in the best possible O(n^2.5), so just keep it in mind.
-fn worley_noise(    density_as_str: &str,
-    positive_threshold_as_str: &str,
-    width_as_str: &str,
-    height_as_str: &str,
-) -> Result<String> {
-    let density = density_as_str.parse::<f64>()?; // density of noise, 0 means no nodes and 100 means that every tile has a node.
-    let positive_threshold = positive_threshold_as_str.parse::<f64>()?; // threshold, if value in cell is above this it gets set to true, otherwise false.
-    let width = width_as_str.parse::<usize>()?;
-    let height = height_as_str.parse::<usize>()?;
+// This is a quite complex algorithm basically what it does is it creates 2 maps, one filled with cells and the other with 'regions' that map onto these cells.
+// Each region can spawn 1 node, the cell then determines wether it is true or false depending on the distance from it to the nearest node in the region minus the second closest node.
+// If this distance is greater than the threshold then the cell is true, otherwise it is false.
+fn worley_noise(str_reg_size : &str, str_positive_threshold: &str, str_width: &str, str_height: &str) -> Result<String>{
 
-    let mut rng = rand::thread_rng();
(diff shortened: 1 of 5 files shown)


AMonkeyThatCodes:
what are u gonna use this for?

EdgeLordExe:
> 
> 
> what are u gonna use this for?

jungle

AMonkeyThatCodes:
biomes divison or something?

EdgeLordExe:
> 
> 
> biomes divison or something?

perlin noise handles biome divisions, gonna use this bad boy for dense foliage generation as this noise provides much more 'forest like' pattern

src/worleynoise.rs
```diff
@@ -0,0 +1,81 @@
...
+    worley_noise(density, threshold, width, height).ok()
+} }
+
+//In the worst possible situation this is O(n^3) in the best possible O(n^2.5), so just keep it in mind.
+fn worley_noise(    density_as_str: &str,
+    positive_threshold_as_str: &str,
+    width_as_str: &str,
+    height_as_str: &str,
+) -> Result<String> {
+    let density = density_as_str.parse::<f64>().unwrap(); // density of noise, 0 means no nodes and 100 means that every tile has a node.
```
willox, line 15: Can you replace all these unwraps with similar error handling to the other noise functions?

https://github.com/tgstation/rust-g/blob/c5cb44d3adc533951b8ea78638d65d74d88c0e45/src/noise_gen.rs#L19
https://github.com/tgstation/rust-g/blob/c5cb44d3adc533951b8ea78638d65d74d88c0e45/src/cellularnoise.rs#L17

It's still not a particularly good approach but in BYOND land nobody knows what to do when stuff crashes, so this works out better at least.

EdgeLordExe: yes sure

willox:
Is this based off of an implementation of Worley noise you've seen somewhere else? I can't find any that look similar and I was hoping to compare them.

EdgeLordExe:
> 
> 
> Is this based off of an implementation of Worley noise you've seen somewhere else? I can't find any that look similar and I was hoping to compare them.

i went off 100% of what i read on multiple sites that documented what worley noise actually is and how it is achieved, I didnt actually read code snippets from a pre-existing implementation. Chances are high there is a faster implementation somewhere that im not aware of.

willox:
Well... I was mostly concerned by the vector that gets sorted once for every single map tile. If you've tested that it's fast enough for what you want to do... I guess I can ignore the performance.

EdgeLordExe:
> 
> 
> Well... I was mostly concerned by the vector that gets sorted once for every single map tile. If you've tested that it's fast enough for what you want to do... I guess I can ignore the performance.

I mean i dont really know any other solution as worley noise requires me to find nth (in this case 1st and 2nd) closest point in every polling point, so unless there is some rust-g thing i dont know about that'd speed it up i'd gladly implement it

EdgeLordExe:
tho there could possibly be performance increase from parallelizing some parts
as individual polling points are independent of each other, same with nodes

willox:
It's fairly slow at larger resolutions (especially over 128x128 and assuming the density isn't set super low, at least.) What sort of map sizes are you planning on passing into this and with what density?

Other implementations seem to work quite differently, and the sorting is definitely the majority of the processing time here.

If you can just confirm that it's actually going to work for what you need I'll merge it.

EdgeLordExe:
> 
> 
> It's fairly slow at larger resolutions (especially over 128x128 and assuming the density isn't set super low, at least.) What sort of map sizes are you planning on passing into this and with what density?
> 
> Other implementations seem to work quite differently, and the sorting is definitely the majority of the processing time here.
> 
> If you can just confirm that it's actually going to work for what you need I'll merge it.

im working on parallelizing it, will look into diffrent solutions

EdgeLordExe:
> 
> 
> It's fairly slow at larger resolutions (especially over 128x128 and assuming the density isn't set super low, at least.) What sort of map sizes are you planning on passing into this and with what density?
> 
> Other implementations seem to work quite differently, and the sorting is definitely the majority of the processing time here.
> 
> If you can just confirm that it's actually going to work for what you need I'll merge it.

i absolutely reworked how this noise works so it is very very fast compared to the previous implementation

src/worleynoise.rs
```diff
@@ -0,0 +1,205 @@
...
+                let cell =  &self.cell_map[i as usize][j as usize];
+                let region = &self.region_map[cell.region.as_ref().reg_x as usize][cell.region.as_ref().reg_y as usize];
+                let neighbours = self.get_regions_in_bound(region.reg_x,region.reg_y,3);
+
+
+                let mut node_vec = Vec::new();
+                node_vec.push(region.as_ref().node.as_ref().unwrap());
+                for neighbour in neighbours {
+                    let node = neighbour.node.as_ref().unwrap();
+                    node_vec.push(node);
```
willox, line 129: Can this end up inserting duplicate references into `node_vec`? Would that even be a problem?

Overall it looks like a better approach. I can look properly tomorrow.

EdgeLordExe: fixed

willox:
As discussed in coderbus, it's not the most idiomatic Rust code but it's self contained and there's no real reason not to let it in.

PR #76, opened 2021-10-02
Adds TOTP generator to the hash feature

adamsong:
Allows for getting TOTP key from key, first time coding in rust so the code is probably bad, accepting any suggestions for how to make it less bad.

### DIFF
--- a/src/hash.rs
+++ b/src/hash.rs
@@ -5,11 +5,11 @@ use md5::Md5;
 use sha1::Sha1;
 use sha2::{Digest, Sha256, Sha512};
 use std::{
+    convert::TryInto,
     fs::File,
     hash::Hasher,
     io::{BufReader, Read},
     time::{SystemTime, UNIX_EPOCH},
-    convert::TryInto,
 };
 use twox_hash::XxHash64;
 
@@ -66,9 +66,7 @@ fn hash_algorithm<B: AsRef<[u8]>>(name: &str, bytes: B) -> Result<String> {
             hasher.write(bytes.as_ref());
             Ok(format!("{:x}", hasher.finish()))
         }
-        "base64" => {
-            Ok(base64::encode(bytes.as_ref()))
-        }
+        "base64" => Ok(base64::encode(bytes.as_ref())),
         _ => Err(Error::InvalidAlgorithm),
     }
 }
@@ -87,7 +85,11 @@ fn file_hash(algorithm: &str, path: &str) -> Result<String> {
 
 /// Generates multiple TOTP codes from 20 character hex_seed, with time step +-tolerance
 /// time_override is used as the current unix time instead of the current system time for testing
-fn totp_generate_tolerance(hex_seed: &str, tolerance: i32, time_override: Option<i64>) -> Result<String> {
+fn totp_generate_tolerance(
+    hex_seed: &str,
+    tolerance: i32,
+    time_override: Option<i64>,
+) -> Result<String> {
     let mut results: Vec<String> = Vec::new();
     for i in -tolerance..(tolerance + 1) {
         let result = totp_generate(hex_seed, i.try_into().unwrap(), time_override)?;
@@ -101,19 +103,25 @@ fn totp_generate_tolerance(hex_seed: &str, tolerance: i32, time_override: Option
 /// TOTP algorithm described https://blogs.unimelb.edu.au/sciencecommunication/2021/09/30/totp/
 /// HMAC algorithm described https://csrc.nist.gov/csrc/media/publications/fips/198/1/final/documents/fips-198-1_final.pdf
 fn totp_generate(hex_seed: &str, offset: i64, time_override: Option<i64>) -> Result<String> {
-
     let mut seed: [u8; 64] = [0; 64];
 
     match hex::decode_to_slice(hex_seed, &mut seed[..10] as &mut [u8]) {
         Ok(value) => value,
-        Err(_) => return Err(Error::HexDecodeError)
+        Err(_) => return Err(Error::HexDecodeError),
     };
 
     let ipad: [u8; 64] = seed.map(|x| x ^ 0x36); // HMAC Step 4
     let opad: [u8; 64] = seed.map(|x| x ^ 0x5C); // HMAC Step 7
 
     // Will panic if the date is not between Jan 1 1970 and the year ~200 billion
-    let curr_time: i64 = time_override.unwrap_or(SystemTime::now().duration_since(UNIX_EPOCH).expect("SystemTime is before Unix Epoc").as_secs().try_into().unwrap()) / 30;
+    let curr_time: i64 = time_override.unwrap_or(
+        SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("SystemTime is before Unix Epoc")
+            .as_secs()
+            .try_into()
+            .unwrap(),
+    ) / 30;
     let time: u64 = (curr_time + offset) as u64;
 
     let time_bytes: [u8; 8] = time.to_be_bytes();
@@ -132,7 +140,7 @@ fn totp_generate(hex_seed: &str, offset: i64, time_override: Option<i64>) -> Res
 
     let offset: usize = (hmac[19] & 0x0F).into();
 
-    let result_bytes: [u8; 4] = hmac[offset..(offset+4)].try_into().unwrap();
+    let result_bytes: [u8; 4] = hmac[offset..(offset + 4)].try_into().unwrap();
 
     let full_result: u32 = u32::from_be_bytes(result_bytes);
     let result: u32 = (full_result & 0x7FFFFFFF) % 1000000;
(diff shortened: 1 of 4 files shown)


willox:
You might want a way to generate a couple of tokens around the current one to add a little leniency for clients with sightly out of sync system times.

src/hash.rs
```diff
@@ -64,3 +70,65 @@ fn file_hash(algorithm: &str, path: &str) -> Result<String> {
...
+
+    let hex_seed_bytes = (0..hex_seed.len())
+                        .step_by(2)
+                        .map(|i| u8::from_str_radix(&hex_seed[i..i + 2], 16));
+
+    let mut i = 0;
+    for b in hex_seed_bytes {
+        seed[i] = b.unwrap();
+        i += 1;
+    }
```
willox, line 85: We already use the [hex](https://crates.io/crates/hex) crate in this module which could decode this input directly into to your array with `hex::decode_to_slice`.

src/hash.rs
```diff
@@ -64,3 +70,65 @@ fn file_hash(algorithm: &str, path: &str) -> Result<String> {
...
+
+    let mut ipad: [u8; 64] = [0; 64];
+    let mut opad: [u8; 64] = [0; 64];
+
+    for j in 0..64 {
+        ipad[j] = seed[j] ^ 0x36;
+        opad[j] = seed[j] ^ 0x5C;
+    }
```
willox, line 93: Since Rust 1.55.0 (which our CI is using) you can replace the loop and declarations here with:
```rust
let mut ipad = seed.map(|x| x ^ 0x36);
let mut opad = seed.map(|x| x ^ 0x5C);
```

I'm also not sure these need to be mutable.

src/hash.rs
```diff
@@ -64,3 +70,65 @@ fn file_hash(algorithm: &str, path: &str) -> Result<String> {
...
+
+    let mut ipad_time_hash_opad: [u8; 84] = [0; 84];
+
+    for j in 0..84 {
+        if j < 64 {
+            ipad_time_hash_opad[j] = opad[j]
+        } else {
+            ipad_time_hash_opad[j] = ipad_time_hash[j - 64]
+        }
+    }
```
willox, line 120: Same story as the rest really. I think you can just get rid of this entire array (and through doing so get rid of the loop.)

willox (changes requested):

src/hash.rs
```diff
@@ -64,3 +70,65 @@ fn file_hash(algorithm: &str, path: &str) -> Result<String> {
...
+
+    let mut ipad_time: [u8; 72] = [0; 72];
+
+    for j in 0..72 {
+        if j < 64 {
+            ipad_time[j] = ipad[j];
+        } else {
+            ipad_time[j] = time_bytes[j - 64];
+        }
+    }
```
willox, line 106: You _could_ get rid of this loop by using `slice::clone_from_slice` like so:

```rust
    let mut ipad_time: [u8; 72] = [0; 72];
    ipad_time[..64].clone_from_slice(&ipad[..]);
    ipad_time[64..].clone_from_slice(&time_bytes[..]);
```

**But**, I think this ipad_time array (and hence the copy operations) shouldn't be necessary at all. You can pass the data into the Sha1 hasher using multiple calls to `Sha1::update`.

Here's an example of that in the Rust playground. It's using Sha2 because that's all that is available there: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4ab98d46d41b57af58e2d6c740936149

willox (changes requested):

adamsong:
The code definitely looks cleaner now, thanks for the help

src/hash.rs
```diff
@@ -64,3 +74,51 @@ fn file_hash(algorithm: &str, path: &str) -> Result<String> {
...
+        println!("Result is now {}", result)
+    }
+    Ok(result.to_string())
+}
+
+fn totp_generate(hex_seed: &str, offset: i64) -> Result<String> {
+    let mut seed: [u8; 64] = [0; 64];
+    let mut seed_bytes = [0u8; 10];
+
+    hex::decode_to_slice(hex_seed, &mut seed_bytes as &mut [u8]).unwrap();
```
willox, line 94: Can't this decode directly into `seed`? I'm not seeing the reason to have both seed and seed_bytes.

src/hash.rs
```diff
@@ -64,3 +74,51 @@ fn file_hash(algorithm: &str, path: &str) -> Result<String> {
 
     Ok(hash_algorithm(algorithm, &bytes)?)
 }
+
+fn totp_generate_tolerance(hex_seed: &str, tolerance: i32) -> Result<String> {
+    let mut result: String = "".to_string();
+    for i in -tolerance..(tolerance + 1) {
+        if i != -tolerance {
+            result.push(',');
```
willox, line 82: Perhaps should output valid JSON instead. I think we have some other functions that do this?

src/hash.rs
```diff
@@ -64,3 +74,51 @@ fn file_hash(algorithm: &str, path: &str) -> Result<String> {
...
+        }
+        result += &totp_generate(hex_seed, i.try_into().unwrap()).unwrap();
+        println!("Result is now {}", result)
```
willox, line 85: Probably a debug message that needs removing?

willox (changes requested):
Yep the code is no longer insanely intimidating at first glance. This is a great improvement.

src/hash.rs
adamsong: Seed is too long, decode_to_slice panics due to incorrect string length

willox: Could you pass a sub-slice of seed that is the correct length?

adamsong: That worked

adamsong:
I left a test in there, I can remove it before this gets merged if you want me to though

willox:
Can you add some sort of error handling? rust-g isn't great at this in general, but crashing on incorrect seed length seems a little too volatile

adamsong:
I will see what I can do, is there an easy way to return an error to byond?

willox:
Nope. All you've got to work with is strings.

You could use JSON again or just settle with something like returning an empty string on error.

adamsong:
Would replacing most of the `unwrap` calls with `?` work?

willox:
Yes. You might have to fiddle with the return value to get that to work though.

Returning None from inside your `byond_fn!` usage will result in an empty string being returned to BYOND. The `.ok()` you've got in there will turn an error into None.

adamsong:
Some of these errors don't like being turned into `Error`, and I cannot set the error type on the result because `Result` has been redefined in `error.rs` to not take a second type parameter. I have a workaround that hopefully works, and some of the panics I am going to leave (If your system clock is past the year 200 billion, that is on you)

src/hash.rs
```diff
@@ -64,3 +84,80 @@ fn file_hash(algorithm: &str, path: &str) -> Result<String> {
...
     Ok(hash_algorithm(algorithm, &bytes)?)
 }
+
+/// Generates multiple TOTP codes from 20 character hex_seed, with time step +-tolerance
+/// time_override is used as the current unix time instead of the current system time for testing
+fn totp_generate_tolerance(hex_seed: &str, tolerance: i32, time_override: Option<i64>) -> Result<String> {
+    let mut results: Vec<String> = Vec::new();
+    for i in -tolerance..(tolerance + 1) {
+        let result = totp_generate(hex_seed, i.try_into().unwrap(), time_override)?;
+        if result.starts_with("ERROR:") {
```
willox, line 94: You definitely want code like this to be closer to:

```rust
let result = totp_generate(hex_seed, i.try_into().unwrap(), time_override)?
results.push(result);
```

Which I assume is what you were actually going for before being blocked by the functions' Result type.

willox (changes requested):
I'm not a fan of this approach.

I've had a look at what's going on in error.rs and it shouldn't be too hard to leverage. First of all just add some error messages into its Error enum here: https://github.com/tgstation/rust-g/blob/539c948a5132c3617045026f2fedfaeb23d5c99a/src/error.rs#L20-L58

Let's add:
```rust
    #[cfg(feature = "hash")]
    #[error(transparent)]
    InvalidSeed,
```

You can do the following to map errors from third-party APIs to your own:
```rust
fn totp_generate(hex_seed: &str, offset: i64, time_override: Option<i64>) -> Result<String> {
    let decode_result = hex::decode_to_slice(hex_seed, &mut seed[..10] as &mut [u8])
        .map_err(|_| Error::InvalidSeed)?;

    ...
}
```

There are also ways to automatically translate between errors of one type to another but I like this approach.

src/hash.rs
adamsong: The problem is I couldn't return a `FromHexError` and when I tried to instantiate `Error` I got an error I didn't understand, so I'm just returning an Ok, with the error message as the text.

adamsong: If I instead try to return `Err(String)` I get 'expected enum `error::Error`' and if I try `Err(Error::new)` I get 'variant or associated item cannot be called on `error::Error` due to unsatisfied trait bounds'

willox: With `Error::new` I assume you were trying to create an instance of the Error enum in our error.rs module? It's an enum so to create an instance you can pick one of the variants and use it directly.

For example our first variant is Null: https://github.com/tgstation/rust-g/blob/539c948a5132c3617045026f2fedfaeb23d5c99a/src/error.rs#L19-L22

To use that in some code you could do:
```rust
fn my_func() -> Result<u32> {
    return Err(Error::Null)
}
```

adamsong: Can I add my own error to the Enum? None of them fit very well

adamsong: I now see that is what you suggested in the original comment

willox (approved):

willox:
I probably won't merge this before tomorrow.

PR #77, opened 2021-11-02
Functions for more precise time measurement

pali6:
Adds `time_reset`, `time_microseconds` and `time_milliseconds` BYOND-accessible functions which can be used to create and reset multiple "timers" and track time since they were last reset more precisely than it is possible in BYOND itself. I personally mostly intend to use these for some detailed live profiling in some performance-problematic parts of https://github.com/goonstation/goonstation but maybe other people can use these too.

I tested that these work but yell at me if I forgot anything or screwed up as Rust is not the language I use most of the time, thanks.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,7 @@ dmsort = {version = "1.0.0", optional = true}
 toml-dep = { version = "0.5.8", package="toml", optional = true }
 
 [features]
-default = ["cellularnoise", "dmi", "file", "git", "http", "json", "log", "noise", "sql", "toml", "url", "time"]
+default = ["cellularnoise", "dmi", "file", "git", "http", "json", "log", "noise", "sql", "time", "toml", "url"]
 
 # default features
 cellularnoise = ["rand"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,7 @@ dmsort = {version = "1.0.0", optional = true}
 toml-dep = { version = "0.5.8", package="toml", optional = true }
 
 [features]
-default = ["cellularnoise", "dmi", "file", "git", "http", "json", "log", "noise", "sql", "toml", "url"]
+default = ["cellularnoise", "dmi", "file", "git", "http", "json", "log", "noise", "sql", "toml", "url", "time"]
 
 # default features
 cellularnoise = ["rand"]
@@ -58,6 +58,7 @@ log = ["chrono"]
 sql = ["mysql", "serde", "serde_json", "once_cell", "dashmap", "jobs"]
 toml = ["serde", "serde_json", "toml-dep"]
 url = ["url-dep", "percent-encoding"]
+time = []
 
 # non-default features
 hash = ["base64", "const-random", "md-5", "hex", "sha-1", "sha2",  "twox-hash", "serde", "serde_json"]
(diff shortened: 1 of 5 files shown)


src/time.rs
```diff
@@ -0,0 +1,39 @@
+use std::{
+    cell::RefCell,
+    collections::hash_map::{Entry, HashMap},
+    ffi::OsString,
+    time::Instant,
+};
+
+thread_local!( static INSTANTS: RefCell<HashMap<OsString, Instant>> = RefCell::new(HashMap::new()) );
```
SpaceManiac, line 8: Using OsString here is likely unnecessary; BYOND has also standardized on UTF-8, just like Rust Strings.

SpaceManiac:
With regard to profiling, [be aware](https://doc.rust-lang.org/std/time/struct.Instant.html):
> Note, however, that instants are not guaranteed to be steady. In other words, each tick of the underlying clock might not be the same length (e.g. some seconds may be longer than others). An instant may jump forwards or experience time dilation (slow down or speed up), but it will never go backwards.

However, the effect is likely to be small.

src/time.rs
pali6: Ah, honestly I just loked at whatever `log.rs` used I think because I wasn't sure.

pali6:
Hm, I'm personally fine with accepting the risk of unsteady time and dunno what else to use. Does `time::SteadyTime` have comparable precision and would it be a better fit?

SpaceManiac:
Considering the `time` crate removed `SteadyTime` in v0.3, they must not have liked it very much, but I don't have any details for or against it myself.

pali6:
Oh lol, I must have had an old version of the docs opened up.

ZeWaka:
@tgstation/rust-g-dev-team could we get this reviewed?

src/lib.rs
```diff
@@ -34,6 +34,8 @@ pub mod unzip;
 pub mod url;
 #[cfg(feature = "worleynoise")]
 pub mod worleynoise;
+#[cfg(feature = "time")]
+pub mod time;
```
Mothblocks, line 38: i can't auto suggest it, but put this in alphabetic order like the rest of them

Cargo.toml
```diff
@@ -58,6 +58,7 @@ log = ["chrono"]
 sql = ["mysql", "serde", "serde_json", "once_cell", "dashmap", "jobs"]
 toml = ["serde", "serde_json", "toml-dep"]
 url = ["url-dep", "percent-encoding"]
+time = []
```
Mothblocks, line 61: abc order

dmsrc/time.dm
```diff
@@ -0,0 +1,3 @@
+#define rustg_time_microseconds(id) call(RUST_G, "time_microseconds")(id)
+#define rustg_time_milliseconds(id) call(RUST_G, "time_milliseconds")(id)
```
Mothblocks, line 2: You should maybe have these call `text2num` if they're going to return as strings

Mothblocks (changes requested):

AffectedArc07:
>  I personally mostly intend to use these for some detailed live profiling in some performance-problematic parts of https://github.com/goonstation/goonstation 

This is just btime 2021 :P

pali6:
@tgstation/rust-g-dev-team are there any further issues with this PR?

optimumtact:
@tgstation/rust-g-dev-team 
Will merge pending no further objections in 24 hours.

PR #80, opened 2021-12-22
Adds Redis Pub/Sub integration

MCHSL:
Adds the ability to connect to Redis, subscribe to messages and publish them as well. [See here for example usage.](https://github.com/ParadiseSS13/Paradise/pull/17266)

The API, prefixed with `rustg_redis_`:
- `connect(addr)` - Connects to a Redis instance using the given address, for example `redis://127.0.0.1/`. Returns an empty string on success, returns the error otherwise.
- `disconnect()` - Closes the connection to Redis and stops the thread managing it. Call this before restarting or attempting to reconnect after an error.
- `subscribe(channel)` - Subscribes to a given channel and starts receiving messages from it.
- `get_messages()` - Returns all received messages as a JSON string, in the format of
    `{"channel_1": ["msg1", "msg2", "msg3", ...], "channel_2": ["msg1", "msg2", "msg3", ...]}`.
    Also includes errors, which appear on the channel `"RUSTG_REDIS_ERROR_CHANNEL"`.
- `publish(channel, message)` - Publishes a message on the given channel.

Remember to check the error channel every time you call `get_messages()`. If any occur, you need to call `disconnect()`, then `connect()` again, then resubscribe to desired channels.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "rust-g"
 edition = "2018"
-version = "0.5.0"
+version = "0.6.0"
 authors = ["Bjorn Neergaard <bjorn@neersighted.com>"]
 repository = "https://github.com/tgstation/rust-g"
 license-file = "LICENSE"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "rust-g"
 edition = "2018"
-version = "0.6.0"
+version = "0.5.0"
 authors = ["Bjorn Neergaard <bjorn@neersighted.com>"]
 repository = "https://github.com/tgstation/rust-g"
 license-file = "LICENSE"
@@ -32,6 +32,7 @@ png = { version = "0.16", optional = true }
 image = { version = "0.23.10", optional = true }
 git2 = { version = "0.13", optional = true, default-features = false }
 noise = { version = "0.7", optional = true}
+redis = { version = "0.21.4", optional = true }
 reqwest = { version = "0.11", optional = true, default-features = false, features = ["blocking", "rustls-tls"] }
 serde = { version = "1.0", optional = true, features = ["derive"] }
 serde_json = { version = "1.0", optional = true }
@@ -43,7 +44,6 @@ zip = { version = "0.5.8", optional = true }
 rand = {version = "0.8", optional = true}
 dmsort = {version = "1.0.0", optional = true }
 toml-dep = { version = "0.5.8", package="toml", optional = true }
-redis = { version = "0.21.4", optional = true }
 
 [features]
 default = ["cellularnoise", "dmi", "file", "git", "http", "json", "log", "noise", "sql", "time", "toml", "url"]
@@ -63,9 +63,9 @@ url = ["url-dep", "percent-encoding"]
 
 # non-default features
 hash = ["base64", "const-random", "md-5", "hex", "sha-1", "sha2",  "twox-hash", "serde", "serde_json"]
+redis_pubsub = ["flume", "redis", "serde", "serde_json"]
 unzip = ["zip", "jobs"]
 worleynoise = ["rand","dmsort"]
-redis_pubsub = ["flume", "redis", "serde", "serde_json"]
 
 # internal feature-like things
 jobs = ["flume"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,8 +41,9 @@ mysql = { version = "20.0", optional = true }
 dashmap = { version = "4.0", optional = true }
 zip = { version = "0.5.8", optional = true }
 rand = {version = "0.8", optional = true}
-dmsort = {version = "1.0.0", optional = true}
+dmsort = {version = "1.0.0", optional = true }
 toml-dep = { version = "0.5.8", package="toml", optional = true }
+redis = { version = "0.21.4", optional = true }
 
 [features]
 default = ["cellularnoise", "dmi", "file", "git", "http", "json", "log", "noise", "sql", "time", "toml", "url"]
@@ -64,6 +65,7 @@ url = ["url-dep", "percent-encoding"]
 hash = ["base64", "const-random", "md-5", "hex", "sha-1", "sha2",  "twox-hash", "serde", "serde_json"]
 unzip = ["zip", "jobs"]
 worleynoise = ["rand","dmsort"]
+redis_pubsub = ["flume", "redis", "serde", "serde_json"]
 
 # internal feature-like things
 jobs = ["flume"]
--- a/README.md
+++ b/README.md
@@ -93,10 +93,10 @@ The default features are:
 
 Additional features are:
 * hash: Faster replacement for `md5`, support for SHA-1, SHA-256, and SHA-512. Requires OpenSSL on Linux.
+* redis_pubsub: Library for sending and receiving messages through Redis.
 * url: Faster replacements for `url_encode` and `url_decode`.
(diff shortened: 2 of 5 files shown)


AffectedArc07:
PR is WIP while I stability test it on para prod

AnturK:
Shouldn't these channels be bounded, then i guess if you subscribe to anything you'll be checking with some regularity.

AffectedArc07:
> Shouldn't these channels be bounded, then i guess if you subscribe to anything you'll be checking with some regularity.

Bounded? In what way?

Cargo.toml
```diff
@@ -41,8 +41,9 @@ mysql = { version = "20.0", optional = true }
 dashmap = { version = "4.0", optional = true }
 zip = { version = "0.5.8", optional = true }
 rand = {version = "0.8", optional = true}
-dmsort = {version = "1.0.0", optional = true}
+dmsort = {version = "1.0.0", optional = true }
```
AffectedArc07, line 44: Also you might want to bump the version of the package itself
`version = "0.5.0"`

I guess with semver this would be `version = "0.6.0"`?

MCHSL:
> Bounded? In what way?

Bounded in the sense of only allowing up to some number of messages waiting in queue, right now if you don't consume them, they can pile up forever until you run out of memory. I s'pose I can cap it to 1000 messages or something

Cargo.toml
ZeWaka: why would you include the patch version

optimumtact:
great, i386 pkgs are broken again

AffectedArc07:
> great, i386 pkgs are broken again

https://github.com/ParadiseSS13/rust-g/pull/1

Hacky but fixes it

optimumtact:
surely that's needed for building the mariadb client lib

AffectedArc07:
> surely that's needed for building the mariadb client lib

It used to be needed for OpenSSL but isn't anymore. CI passed without pkg-config even when building with all features.

AffectedArc07:
Seems to be stable on para.

Cargo.toml
AffectedArc07: > why would you include the patch version

I assumed that was standard with semver. RUSTG has always been 0.x.0

willox:
pls merge/rebase with master

MCHSL:
Changed it so if the receive queue is full, any further received messages are immediately dropped. Otherwise, it'd prevent sending, subscribing, and stopping from working, and could even deadlock if you attempted to send more than 1k messages while it was blocked.

MCHSL:
Can now handle calling connect several times and forgetting to call disconnect before connecting again

Cargo.toml
```diff
@@ -64,6 +65,7 @@ url = ["url-dep", "percent-encoding"]
 hash = ["base64", "const-random", "md-5", "hex", "sha-1", "sha2",  "twox-hash", "serde", "serde_json"]
 unzip = ["zip", "jobs"]
 worleynoise = ["rand","dmsort"]
+redis_pubsub = ["flume", "redis", "serde", "serde_json"]
```
Mothblocks, line 68: abc order

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "rust-g"
 edition = "2018"
-version = "0.5.0"
+version = "0.6.0"
```
Mothblocks, line 4: Don't bump the version here--we'll get to it in its own PR

README.md
```diff
@@ -95,7 +95,8 @@ Additional features are:
 * hash: Faster replacement for `md5`, support for SHA-1, SHA-256, and SHA-512. Requires OpenSSL on Linux.
 * url: Faster replacements for `url_encode` and `url_decode`.
 * unzip: Function to download a .zip from a URL and unzip it to a directory.
-* worleynoise: Function that generates a type of nice looking cellular noise, more expensive than cellularnoise
+* worleynoise: Function that generates a type of nice looking cellular noise, more expensive than cellularnoise.
+* redis_pubsub: Library for sending and receiving messages through Redis.
```
Mothblocks, line 99: ABC order

src/redis_pubsub.rs
```diff
@@ -0,0 +1,165 @@
...
+    }
+}
+
+fn handle_redis(
+    client: Client,
+    control: flume::Receiver<PubSubRequest>,
+    out: flume::Sender<PubSubResponse>,
+) {
+    if let Err(e) = handle_redis_inner(client, &control, &out) {
+        let _ = out.try_send(PubSubResponse::Disconnected(e.to_string()));
```
Mothblocks, line 78: Why are we silencing potential errors here? This is spooky.

src/redis_pubsub.rs
```diff
@@ -0,0 +1,165 @@
...
+
+fn handle_redis_inner(
+    client: Client,
+    control: &flume::Receiver<PubSubRequest>,
+    out: &flume::Sender<PubSubResponse>,
+) -> Result<(), RedisError> {
+    let mut conn = client.get_connection()?;
+    let mut pub_conn = client.get_connection()?;
+    let mut pubsub = conn.as_pubsub();
+    let _ = pubsub.set_read_timeout(Some(Duration::from_secs(1)));
```
Mothblocks, line 33: I'm assuming this is another `Result` `must_use`, what's the error states of this and why are we avoiding it?

src/redis_pubsub.rs
```diff
@@ -0,0 +1,165 @@
...
+    });
+    RESPONSE_RECEIVER.with(|cell| {
+        cell.replace(None);
+    });
+}
+
+fn subscribe(channel: &str) {
+    REQUEST_SENDER.with(|cell| {
+        if let Some(chan) = cell.borrow_mut().as_ref() {
+            let _ = chan.send(PubSubRequest::Subscribe(channel.to_owned()));
```
Mothblocks, line 106: Same

src/redis_pubsub.rs
```diff
@@ -0,0 +1,165 @@
...
+        };
+    });
+}
+
+fn publish(channel: &str, msg: &str) {
+    REQUEST_SENDER.with(|cell| {
+        if let Some(chan) = cell.borrow_mut().as_ref() {
+            let _ = chan.send(PubSubRequest::Publish(channel.to_owned(), msg.to_owned()));
```
Mothblocks, line 114: Same

Mothblocks (changes requested):

AffectedArc07:
Once changes are addressed please do not immediately merge, as I need to stability test it on prod.

Mothblocks:
Draft then plz

src/redis_pubsub.rs
MCHSL: This could fail if the receiving end was dropped, in which case there's no need (or way) to communicate back the error, or if the channel was full, in which case the DM side was probably FUBAR. I suppose it doesn't hurt to make it block until there's space or if the receiver is dropped.

AffectedArc07: If this blocks, how is the queue ever going to clear if DM cant execute anything to clear the queue due to being blocked by this?

MCHSL:
subscribe and publish now also return errors, they can stem either from failing to connect or being unable to keep up with incoming messages and letting buffers fill up

src/redis_pubsub.rs
MCHSL: This runs on the redis thread side, it won't block DM

AffectedArc07:
TM'd on paradise prod for about a day straight with no crashes and decent levels of activity (relaying all msay and asay messages to and from discord)

PR #83, opened 2021-12-26
Cleanups and additions for the Aho-Corasick replacement stuff

vuonojenmustaturska:
Most notably adds an option to make case-insensitive replacements. Other options are provided as a courtesy. Fiddly DFA/NFA settings are not provided as auto_configure automagically sets up DFA if the matches array has less than 100 items.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ cargo build --release --target i686-pc-windows-msvc
 To get additional features, pass a list to `--features`, for example `--features hash,url`. To get all features, pass `--all-features`. To disable the default features, pass `--no-default-features`.
 
 The default features are:
+* acreplace: Aho-Corasick string matching and replacement.
 * cellularnoise: Function to generate cellular automata-based noise.
 * dmi: DMI manipulations which are impossible from within BYOND.
   Used by the asset cache subsystem to improve load times.
@@ -86,11 +87,10 @@ The default features are:
 * http: Asynchronous HTTP(s) client supporting most standard methods.
 * json: Function to check JSON validity.
 * log: Faster log output.
-* sql: Asynchronous MySQL/MariaDB client library.
 * noise: 2d Perlin noise.
-* toml: TOML parser.
+* sql: Asynchronous MySQL/MariaDB client library.
 * time: High-accuracy time measuring.
-* acreplace: Aho-Corasick string matching and replacement.
+* toml: TOML parser.
 
 Additional features are:
 * hash: Faster replacement for `md5`, support for SHA-1, SHA-256, and SHA-512. Requires OpenSSL on Linux.
--- a/README.md
+++ b/README.md
@@ -90,6 +90,7 @@ The default features are:
 * noise: 2d Perlin noise.
 * toml: TOML parser.
 * time: High-accuracy time measuring.
+* acreplace: Aho-Corasick string matching and replacement.
 
 Additional features are:
 * hash: Faster replacement for `md5`, support for SHA-1, SHA-256, and SHA-512. Requires OpenSSL on Linux.
--- a/src/acreplace.rs
+++ b/src/acreplace.rs
@@ -36,8 +36,8 @@ where
     D: serde::de::Deserializer<'de>,
 {
     match u8::deserialize(deserializer)? {
-        1 => Ok(true),
-        _ => Ok(false),
+        0 => Ok(false),
+        _ => Ok(true),
     }
 }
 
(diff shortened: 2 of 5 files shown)


src/acreplace.rs
```diff
@@ -1,22 +1,64 @@
...
+    pub replacements: Vec<String>,
+}
+
+#[derive(Deserialize)]
+struct AhoCorasickOptions {
+    #[serde(default, deserialize_with = "deserialize_byond_bool")]
+    pub anchored: bool,
+    #[serde(default, deserialize_with = "deserialize_byond_bool")]
+    pub ascii_case_insensitive: bool,
+    #[serde(default, deserialize_with = "deserialize_matchkind")]
```
vuonojenmustaturska, line 19: Not using serde's remote derive here as I feel it looks messy and the underlying crate receiving new `MatchKind`s would cause compile errors in rust-g, whereas with this new `MatchKind`s are merely not supported without someone adding them.

vuonojenmustaturska:
with 5000 calls and `The Syndicate have cunningly disguised a Syndicate Uplink as your PDA-Francis Davies on your belt. Simply enter the code "423 Delta" into the ring message select to unlock its hidden features.` as test input, the benchmarks are:

* default shakespearify: 2.282
* case-sensitive rust-g: 0.333
* case-insensitive rust-g: 0.355
* replacements array supplied with haystack with case-insensitive rust-g: 0.379

Non-default matchkinds may have added benefits.

README.md
```diff
@@ -90,6 +90,7 @@ The default features are:
 * noise: 2d Perlin noise.
 * toml: TOML parser.
 * time: High-accuracy time measuring.
+* acreplace: Aho-Corasick string matching and replacement.
```
Mothblocks, line 93: ABC order (even though NOISE isnt....)

src/acreplace.rs
```diff
@@ -1,22 +1,64 @@
...
+        .build(patterns)
+    }
+}
+
+fn deserialize_byond_bool<'de, D>(deserializer: D) -> Result<bool, D::Error>
+where
+    D: serde::de::Deserializer<'de>,
+{
+    match u8::deserialize(deserializer)? {
+        1 => Ok(true),
```
Mothblocks, line 39: Shouldn't this be the other way around? Any non-zero value is true?

src/acreplace.rs
```diff
@@ -25,14 +67,33 @@ byond_fn! { setup_acreplace(key, patternsjson, replacementsjson) {
...
+
 
 byond_fn! { acreplace(key, text) {
     CREPLACE_MAP.with(|cell| -> Option<String> {
         let map = cell.borrow_mut();
-        match map.get(&key.to_owned()) {
-            Some(replacements) => Some(replacements.automaton.replace_all(text, &replacements.replacements)),
-            None => None
-        }
+        let replacements = map.get(&key.to_owned())?;
```
Mothblocks, line 86: Why `to_owned()` then immediately referencing it?

src/acreplace.rs
```diff
@@ -25,14 +67,33 @@ byond_fn! { setup_acreplace(key, patternsjson, replacementsjson) {
...
+        Some(replacements.automaton.replace_all(text, &replacements.replacements))
     })
 } }
 
+byond_fn! { acreplace_with_replacements(key, text, replacements_json) {
+    let call_replacements: Vec<String> = serde_json::from_str(&replacements_json).ok()?;
+    CREPLACE_MAP.with(|cell| -> Option<String> {
+        let map = cell.borrow_mut();
+        let replacements = map.get(&key.to_owned())?;
```
Mothblocks, line 95: Same

src/acreplace.rs
Mothblocks: Hesitant :+1:--adding new `MatchKind` without `non_exhaustive` is a breaking change anyway, so we shouldn't be just updating entire major versions frivolously.

src/acreplace.rs
vuonojenmustaturska: The default bool value is false so I thought it fitting, but since it's definitely a number at this point it would make more sense indeed

README.md
vuonojenmustaturska: neither was toml

src/acreplace.rs
vuonojenmustaturska: autopiloted to `.to_owned()` at some point in development and then made it work with the ampersand. removed.

PR #85, opened 2021-12-29
Fix clippy lints, Rustfmt, put both on CI, and update Cargo.lock

Mothblocks:
Fixes all Clippy lints, runs `cargo fmt` on everything, requires both on CI, and updates Cargo.lock which was incorrectly not updated with the aho corasick PR.


ZeWaka (approved):
pretty straightforward except for the unsafe block addition

PR #88, opened 2021-12-30
fixes warning in redis

MCHSL:


src/redis_pubsub.rs
```diff
@@ -4,7 +4,7 @@ use std::collections::HashMap;
 use std::thread;
 use std::time::Duration;
 
-static ERROR_CHANNEL: &'static str = "RUSTG_REDIS_ERROR_CHANNEL";
+static ERROR_CHANNEL: &str = "RUSTG_REDIS_ERROR_CHANNEL";
```
Mothblocks, line 7: Can you make this `const` while you're here

PR #89, opened 2021-12-30
Fix clippy CI, format byond_fn

Mothblocks:
Changes action/clippy to just action/cargo so that we don't have to read awful JSON dumps.

Changes .editorconfig to LF mode, because VSC keeps changing it to CRLF, and I don't know why.

Changes `byond_fn!` from `byond_fn! { name() {} }` to `byond_fn!(fn name() {})` so that Rustfmt picks up on it.

Adds `--locked` to Clippy's CI so that we can make sure Cargo.lock is up to date.

Updates Cargo.lock because it kept doing it I guess.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -21,7 +21,7 @@ jobs:
         with:
           toolchain: stable
           command: clippy
-          args: --target i686-pc-windows-msvc --all-features -- -D warnings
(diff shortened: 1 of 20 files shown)


ZeWaka (commented):
not going to add on the `fn` to hash.rs, redis...rs, unzip.rs?

Mothblocks:
I thought I did

Mothblocks:
rust-analyzer SSR probably broke, oh well

PR #94, opened 2022-02-11
feat: use local time

Bizzonium:


ZeWaka:
why dough

optimumtact:
You can't change the default like this, as everyone preexisting is relying on it being UTC.

you have to add a feature tag to enable local time and use your own compiled version.

PR #95, opened 2022-04-25
Add functions for seeking lines and getting line counts in functions

Mothblocks:
Adds `rustg_file_get_line_count` for getting the line count of a given filename, and `rustg_file_seek_line` for getting a specific line.

These are useful as to not load the entire file into memory in DM, which is useful for very large files like dictionaries.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -28,7 +28,7 @@ fn main() {
     for (key, _value) in std::env::vars() {
         // CARGO_FEATURE_<name> — For each activated feature of the package being built, this environment variable will be present where <name> is the name of the feature uppercased and having - translated to _.
         if let Some(uprfeature) = key.strip_prefix("CARGO_FEATURE_") {
-            let feature = uprfeature.to_lowercase().replace("_", "-"); // actual proper name of the enabled feature
+            let feature = uprfeature.to_lowercase().replace('_', "-"); // actual proper name of the enabled feature
             if feature_dm_exists!(&feature) {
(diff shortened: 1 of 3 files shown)


AffectedArc07 (changes requested):
`build.rs` L 31:
```rs
let feature = uprfeature.to_lowercase().replace("_", "-"); // actual proper name of the enabled feature
```
to
```rs
let feature = uprfeature.to_lowercase().replace('_', '-'); // actual proper name of the enabled feature
```

Mothblocks:
clippy moment

Mothblocks:
i'll do it later :)

AffectedArc07:
@Mothblocks  Also, on the topic of this
`These are useful as to not load the entire file into memory in DM, which is useful for very large files like dictionaries.`

Most dictionaries are JSON, so you would have to account for reformatting it to be readable, as well as keeping the lines in code sane to make sure it grabs the right part of the dictionary. 

This seems like more of a hassle than its worth, to be truly honest.

Mothblocks:
There are two use cases that this is made for:

1. @AnturK wanted to have the wordle dictionary, which is just a text file of words.
2. I'm using this for something where, instead of a JSON of arrays, it's just a text file where each thing is a JSON.

Both these dictionaries can be pretty sizable, and BYOND's memory limits without LAA are really small, especially when you remember rust-g's binary size contributes to it.

PR #98, opened 2022-06-02
toml to json now returns error reasons to byond + rust-analyzer.cargo.target gnu > msvc

### LINKED ISSUE #86: Why is the Rust Analyzer target set to i686-pc-windows gnu?
https://github.com/tgstation/rust-g/blob/274fb1836d83836d8b613cf300435d75870d7a1b/.vscode/settings.json#L2

Title, this has bothered contributors besides myself, who have to tell git to ignore our changes to make it the msvc target instead.

tralezab:
toml to json: One half of better error reporting for things like banned word filters in game, the other half has to be displaying the error on DM side.

rust-analyzer.cargo.target: fixes #86 and let me say holy shit that was one hellish part of getting the analyzer working

BLAZING
LY FAST

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,3 @@
 {
-    "rust-analyzer.cargo.target": "i686-pc-windows-gnu"
+    "rust-analyzer.cargo.target": "i686-pc-windows-msvc"
 }
--- a/README.md
+++ b/README.md
@@ -54,6 +54,11 @@ System libraries:
 
 * Other Linux distributions install the appropriate **32-bit development** and **32-bit runtime** packages.
 
+If you want to use the `pc-windows-gnu` or similar other target ABI, do the following:
+1. Change the `"rust-analyzer.cargo.target"` setting in `.cargo/config` to `i686-pc-windows-gnu`.
+2. Run `git update-index --assume-unchanged .cargo/config`, which will tell git to 'ignore' the changes you made.
+3. If you find yourself ever wanting to change back, run `git update-index --no-assume-unchanged .cargo/config`.
+
 ## Compiling
 
 The [Cargo] tool handles compilation, as well as automatically downloading and
--- a/dmsrc/toml.dm
+++ b/dmsrc/toml.dm
@@ -1,1 +1,8 @@
-#define rustg_read_toml_file(path) json_decode(call(RUST_G, "toml_file_to_json")(path) || "null")
+#define rustg_raw_read_toml_file(path) json_decode(call(RUST_G, "toml_file_to_json")(path) || "null")
+
+/proc/rustg_read_toml_file(path)
+	var/list/output = rustg_raw_read_toml_file(path)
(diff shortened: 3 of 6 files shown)


Mothblocks:
Right now this is going to cause a backwards incompatibility issue where `rustg_read_toml_file` does something different now

I think you should change dmsrc to be like this:

```dm
#define rustg_read_toml_file_result(path) json_decode(call(RUST_G, "toml_file_to_json")(path) || "null")

/proc/rustg_read_toml_file(path)
    var/output = rustg_read_toml_file_result(path)
    if (output["success"])
        return output["content"]
    else
        CRASH(output["content"])
```

That way existing consumers and consumers that just want to runtime instead of need to manage it can use the existing thing, and anything that wants to especially handle the error case can use `rustg_read_toml_file_result`

tralezab:
Aye aye, cap'n!

Mothblocks:
@tralezab Needs to be in dmsrc

.vscode/settings.json
```diff
@@ -1,3 +1,3 @@
 {
-    "rust-analyzer.cargo.target": "i686-pc-windows-gnu"
+    "rust-analyzer.cargo.target": "i686-pc-windows-msvc"
```
ZeWaka, line 2: unlikely spacemaniac will accept this, per the issue

Mothblocks: he didnt comment on the issue though

AffectedArc07: This applies to literally everyone else though, and we shouldn't impact the majority for the sake of one person.

AffectedArc07:
> Right now this is going to cause a backwards incompatibility issue where `rustg_read_toml_file` does something different now
> 
> I think you should change dmsrc to be like this:
> 
> ```dm
> #define rustg_read_toml_file_result(path) json_decode(call(RUST_G, "toml_file_to_json")(path) || "null")
> 
> /proc/rustg_read_toml_file(path)
>     var/output = rustg_read_toml_file_result(path)
>     if (output["success"])
>         return output["content"]
>     else
>         CRASH(output["content"])
> ```
> 
> That way existing consumers and consumers that just want to runtime instead of need to manage it can use the existing thing, and anything that wants to especially handle the error case can use `rustg_read_toml_file_result`

Make it `rustg_raw_read_toml_file()`, `result` is questionable.

.vscode/settings.json
tralezab: this made it so much more painful to start guys

tralezab:
👍

tralezab:
I don't know what's up with the linux build...

willox:
Looks like it is encoding the `enabled` field as `1` instead of `true`.

optimumtact:
you need to fix the unit tests to expect the backtrace

/test/proc/check_toml_file2json ------------------------------------------

tralezab:
Now I understand what's up I got it

PR #99, opened 2022-06-11
Adds a new type of noise: Discrete Batched Perlin-like Noise (DBP)

EdgeLordExe:
Hello!, I've recently been working on jungleland for yogstation, and i realized that calling current rust-g perlin noise255x255 per z level is **extremely** slow, so i coded this. DBP is a type of noise i recently coded and published to crates.io, so i decided to add it to rust-g. 
Here is the code to DBP https://github.com/EdgeLordExe/batch-perlin-like-noise-rs

DBP is extremely fucking fast, it's multithreaded and most importantly it allows byond to not waste time calling perlin noise like thousands of times, instead calling only once and recieveing the results in a single string (just how cellularnoise does it)

Some screenshots: 
obraz

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,7 @@ aho-corasick = { version = "0.7.18", optional = true}
 dbpnoise = { version = "0.1.2", optional = true}
 
 [features]
-default = ["acreplace", "cellularnoise", "dmi", "file", "git", "http", "json", "log", "noise", "sql", "time", "toml", "url", "batchnoise"]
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "rust-g"
 edition = "2018"
-version = "0.7.0"
+version = "0.7.1"
```
Mothblocks, line 4: Don't change the version in this PR (also this would be 0.8.0 anyway)

EdgeLordExe: sure

PR #106, opened 2022-07-23
1.0.0

### LINKED ISSUE #100: 1.0.0
I know it doesn't mean much in a context like this, but rust-g has been battle tested in many major servers and likely will not receive many backwards incompatible changes for a bit. Would be nice to move to 1.0.0 to better communicate what is happening in the updates.

ZeWaka:
Resolves #100

Yes, I migrated (via `cargo fix --edition`) and tested (by running goon with the dll) the edition change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "rust-g"
-edition = "2018"
-version = "0.8.0"
+edition = "2021"
+version = "1.0.0"
 authors = ["Bjorn Neergaard <bjorn@neersighted.com>"]
 repository = "https://github.com/tgstation/rust-g"
 license-file = "LICENSE"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,8 +2,9 @@
 name = "rust-g"
 edition = "2021"
 version = "1.0.0"
-authors = ["Bjorn Neergaard <bjorn@neersighted.com>"]
+authors = ["Bjorn Neergaard <bjorn@neersighted.com>, Tad Hardesty <tad@platymuus.com>, rust-g maintainer team"]
 repository = "https://github.com/tgstation/rust-g"
+license = "MIT"
 license-file = "LICENSE"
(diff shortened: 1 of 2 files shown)


optimumtact:
necoarc

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "rust-g"
-edition = "2018"
-version = "0.8.0"
+edition = "2021"
+version = "1.0.0"
 authors = ["Bjorn Neergaard <bjorn@neersighted.com>"]
```
optimumtact, line 5: i feel like we sould update this at some point

ZeWaka: what do you think we should put? Only neer and space could be considered the 'authors', unless we just generically put `/tg/station and friends <idksomeaddress@tgstation.com>` or whatever email you guys use

optimumtact: i meant adding spacemaniac somewhere

is there a maintainers field too? you can put rust-g dev team there

also we have no emails

ZeWaka: > The optional authors field lists people or organizations that are considered the "authors" of the package. The exact meaning is open to interpretation — it may list the original or primary authors, current maintainers, or owners of the package. An optional email address may be included within angled brackets at the end of each author entry.

Cargo.toml
```diff
@@ -1,10 +1,10 @@
 [package]
 name = "rust-g"
-edition = "2018"
-version = "0.8.0"
-authors = ["Bjorn Neergaard <bjorn@neersighted.com>"]
+edition = "2021"
+version = "1.0.0"
+authors = ["Bjorn Neergaard <bjorn@neersighted.com>, Tad Hardesty <tad@platymuus.com>, rust-g maintainer team"]
```
Mothblocks, line 5: shouldn't this have been multiple entries in the authors array

ZeWaka: oh. yes lmao

PR #107, opened 2022-07-24
Pallette Encoding Image Fix

ZeWaka:
Should copy palette information over for the metadata stripping operation now so the encoding doesn't fail. 

<details>
<summary>Rant</summary>
Annoyingly, the `png` changelog didn't hint at this palette issue - though I'm not entirely sure how it was getting the info before. Nor do I think it's the difference between `png::Info` vs `png::OutputInfo`. Probably some internal mechanism for checking it got changed and our produced images had no palette information before.
</details>

Also threw in a zTxt copy mechanism if at some point we want it, because we now have access to that information there and it's possible in the new library version.

Fixes issue brought up in: https://github.com/tgstation/tgstation/pull/68690

### DIFF
--- a/src/dmi.rs
+++ b/src/dmi.rs
@@ -1,5 +1,5 @@
 use crate::error::{Error, Result};
-use png::{Decoder, Encoder, OutputInfo};
+use png::{Decoder, Encoder, OutputInfo, Reader};
 use std::{
     fs::{create_dir_all, File},
     path::Path,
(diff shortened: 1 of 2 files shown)


src/dmi.rs
```diff
@@ -26,24 +26,41 @@ byond_fn!(fn dmi_resize_png(path, width, height, resizetype) {
...
+    Ok((reader, frame_info, buf))
 }
 
-fn write_png(path: &str, info: OutputInfo, image: Vec<u8>) -> Result<()> {
+fn write_png(
+    path: &str,
+    reader: Reader<File>,
+    info: OutputInfo,
+    image: Vec<u8>,
+    strip: bool,
```
Mothblocks, line 46: Doesn't look like you set this to false anywhere. Needed?

ZeWaka: > Also threw in a zTxt copy mechanism if at some point we want it, because we now have access to that information there and it's possible in the new library version.

ZeWaka: also this fix works, will merge tomorrow and cut release
https://github.com/ZeWaka/tgstation/pull/2 (ignore the dumb tramstation random error)

ZeWaka:
also this fix works, will merge tomorrow and cut release
https://github.com/ZeWaka/tgstation/pull/2 (ignore the dumb tramstation random error)

src/dmi.rs
ZeWaka: aka, useful for preserving DMI information

Mothblocks: Oh okay

PR #108, opened 2022-07-25
Uploads rust_g.dm as an Actions Artifact

### LINKED ISSUE #92: CI should generate rust_g.dm on default features
It's really annoying to do this for releases

san7890:
As requested, so delivered. I have skill of noob so I'm going to use Python for this. Basically, every time this workflow runs, you get an artifact of `rust_g.dm`. You can then use that for your releases, you little scamp.

I'm also going to pass through some DM-side changes to make the final product more readable and not goopy at times.

I'm renaming main.dm to _main.dm because I'm lazy and I can not figure out how to do it with shutil or sort or whatever bro i'm tired. It looks better for the generated dmsrc file, and is probably the order you were making manually anyways. I'm going to go bang my head against a wall, let me know if the newline formatting is cringe.

Closes #92.


san7890:
Short example, you can view the generated `rust_g.dm` file for this exact PR here: https://github.com/tgstation/rust-g/actions/runs/2731080015

.github/workflows/generate_dmsrc.yml
```diff
@@ -0,0 +1,45 @@
...
+    - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+        - uses: actions/checkout@v2
+        - uses: actions/setup-python@v2
+        # Generate the actual .dm file.
+        - uses: jannekem/run-python-script-action@v1
```
Mothblocks, line 19: You shouldn't generate it like this.

`cargo build` already creates the .dm file through the build script.

san7890:
Alright, I massively overthought the potential issue. This is apparently very easy to accomplish using upload-artifact/v3 (which I wouldn't have known if I didn't slog through the initial version of this PR), so not all was lost.

You can view the related, successful version of this here: https://github.com/tgstation/rust-g/actions/runs/2731254461

ZeWaka (approved):

optimumtact:
there is no dm file on that link?

san7890:
> there is no dm file on that link?

Github Actions will automatically spit it out as a .ZIP file irregardless of what you do (even if it is just a singular file), the DM file is included in that ZIP file should you download either the DLL or SO. I'm unsure what I could change the name to, since I can see how that is confusing.

san7890:
Newest Example Link: https://github.com/tgstation/rust-g/actions/runs/2735453994

PR #112, opened 2022-07-27
Returns the TOML library to previous functionality

### LINKED ISSUE #111: rustg_read_toml_file is broken
As far as I can tell, #98 broke rustg_read_toml_file.  We are using it in a tgstation branch, and when I happened to update rust_g, everything that depended on toml files broke.

the toml_file_to_json function in the library passes a json string containing "success" and "content" to byond, and the byond wrapper proc returns the value of "content" if it succeeds.  From what I can tell, the value of content is escaped so that it is not actually parsed by json_decode() during this process, and returned as a json-formatted string (thankfully, with escape characters removed) instead of an associative array.

My current workaround is to json_decode the content again before you return it, and that appears to be working.

Test code (in an otherwise empty test environment, with rust_g.dm and a data.toml file to test):

```
world.New()
        world.log << "Starting"
        var/list/result = rustg_read_toml_file("data.toml")
        world.log << "Read done"
        if(istext(result))
                world.log << "It's just text"
                world.log << result
                var/list/result2 = json_decode(result)
                if(istext(result2))
                        world.log << "still just text"
                else
                        world.log << "converted?!?"
```

For me, this code returns the incredulous "converted?!?" text.  The workaround is also functioning on a live server, providing the correct associative list for the data loaded from file.

ZeWaka:
The DM-side of the API returns an object instead of a JSON string

Fixes #111

### DIFF
--- a/tests/dm/toml.dme
+++ b/tests/dm/toml.dme
@@ -17,5 +17,5 @@ var/test_json = @{"
 
     var/toml_output = rustg_read_toml_file("test.toml")
 
-    if (toml_output != test_json)
+    if (toml_output != json_decode(test_json)) // We want to compare the decoded *objects*
         CRASH("test:\n[test_toml]\n \nexpected:\n[test_json]\n \nrustg:\n[toml_output]")
--- a/tests/dm/toml.dme
+++ b/tests/dm/toml.dme
@@ -9,7 +9,7 @@ temp_targets = { cpu = 79, case = 72 }
 "}
 
 var/test_json = @{"
-{"database":{"data":[["delta","phi"]],"enabled":1,"ports":[8000,25565],"temp_targets":{"case":72,"cpu":79}}}
+{"database":{"data":[["delta","phi"]],"enabled":true,"ports":[8000,25565],"temp_targets":{"case":72,"cpu":79}}}
 "}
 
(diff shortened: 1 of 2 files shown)


tests/dm/toml.dme
```diff
@@ -17,5 +17,5 @@ var/test_json = @{"
 
     var/toml_output = rustg_read_toml_file("test.toml")
 
-    if (toml_output != test_json)
+    if (json_decode(toml_output) ~= json_decode(test_json)) // We want to compare the decoded *objects*
```
SuperSayu, line 20: json_decode(object) ~= json_decode(string)

You either want to compare toml_output (object) to json_decode(test_json) (object) or re-encode both of those into json text and compare the strings (Which is how it worked prior to #98)

ZeWaka: yeah it was just midnight and i was extremely tired, like my eyes were involuntarily closing

ZeWaka:
been busy

ZeWaka:
@SuperSayu lmk if that build: https://github.com/tgstation/rust-g/actions/runs/2768554579 works for you

SuperSayu:
I can't test the build directly because we're on linux, I'd have to build it myself, but the change matches my workaround which has been live and working properly

PR #113, opened 2022-08-01
Add an a* pathfinder

BraveMole:
Returns the shortest path in a static node map. That is made for TGMC wich uses manually placed nodes for pathfinding. 

Benchmark : 
image
That's the average number of path computed in one second, out of 30 runs with random nodes. Size of the node map is roughly 800

On average 32 times faster than tgmc a* implementation. 
Another possible comparison is with TG's JPS system, and according to the benchmark done here it's 7000 times faster (https://github.com/tgstation/tgstation/pull/56780). Not exactly the same use cases though


src/pathfinder.rs
```diff
@@ -0,0 +1,129 @@
...
+impl Node {
+    fn successors(&self) -> Vec<(Self, usize)> {
+        self.connected_nodes_id.iter().map(|id| {
+            if let Some(connected_node) = NODES.read().unwrap().get(*id) {
+                Some(connected_node.clone())
+            }
+            else {
+                None
+            }
+        }).filter_map(|node| node).map(|node| (self.distance(&node), node)).map(|(distance, node)| (node.clone(), distance)).collect()
```
optimumtact, line 24: you're sick dude, you're sick

optimumtact: this is literally unreadable

AffectedArc07:
So to be absolutely clear.

This is for **static** maps so things can pathfind through yes?

BraveMole:
> So to be absolutely clear.
> 
> This is for **static** maps so things can pathfind through yes?

Yeah, you need to prenode.

AffectedArc07:
> > So to be absolutely clear.
> 
> > 
> 
> > This is for **static** maps so things can pathfind through yes?
> 
> 
> 
> Yeah, you need to prenode. 

Ok so to ask the incredibly dumb question.

If we need to pre node, why don't we just pre path for even more performance?

BraveMole:
> If we need to pre node, why don't we just pre path for even more performance?

Typical map has 1500 nodes, so that's about 2 millions path to compute and then save. I'd say it's not worth it. 

Plus in TGMC system you can place additional goal nodes that will link to the node map

src/pathfinder.rs
```diff
@@ -0,0 +1,266 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    #[allow(unused_must_use)]
+    fn test_register_node() {
+        let json = std::fs::read_to_string("tests/rsc/ai_nodes_info.json").unwrap();
+        register_nodes_(&json).is_ok();
+        assert!(NODES.read().unwrap().len() != 0)
```
BraveMole, line 245: when asserting register nodes directly, it fails 50% of the time. I don't know why, if anyone has an idea please shoot

src/pathfinder.rs
```diff
@@ -0,0 +1,266 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    #[allow(unused_must_use)]
+    fn test_register_node() {
+        let json = std::fs::read_to_string("tests/rsc/ai_nodes_info.json").unwrap();
+        register_nodes_(&json).is_ok();
```
BraveMole, line 244:
```suggestion
        register_nodes_(&json);
```

src/pathfinder.rs
```diff
@@ -0,0 +1,266 @@
...
+    y: usize,
+    z: usize,
+    connected_nodes_id: Vec<usize>,
+}
+
+impl Node {
+    fn successors(&self) -> Vec<(Self, usize)> {
+        self.connected_nodes_id
+            .iter()
+            .filter_map(|id| NODES.read().unwrap().get(*id).cloned())
```
BraveMole, line 21: By the way, this constant cloning is terrible but I'm not sure how to avoid it. Using rc ?

ZeWaka (changes requested):
would appreciate *any* comments in the file because right now there is not even one

src/pathfinder.rs
```diff
@@ -0,0 +1,356 @@
...
+use pathfinding::prelude::astar;
+use serde::{Deserialize, Serialize};
+use std::borrow::Borrow;
+use std::hash::Hash;
+use std::num::ParseIntError;
+use std::sync::{Arc, RwLock};
+use std::usize::MAX;
+
+// static mut are not thread safe, and manipulating one is only possible behind unsafe tag
+// However, most of the unsafe call should be thread safe because nodes inside this vec are behind a RwLock
```
BraveMole, line 11: Feel free to correct me on that, I have barely any experience in rust and wasn't able to affirm or infirm that

src/pathfinder.rs
```diff
@@ -0,0 +1,356 @@
...
+static mut NODES: Vec<NodeContainer> = Vec::new();
+
+// Container for a node. Exist mainly to be able to implement Hash, which is not implemented for RefCell
```
BraveMole, line 14: I don't really know why Hash isn't implemented for RefCell, anyone knows why ?

BraveMole:
Clippy fails because of external crate... Halp

src/pathfinder.rs
Mothblocks: HashMaps play badly (not UB, but unspecified) if the hashes change while the object is inside. Interior mutability makes that trivial.

src/pathfinder.rs
```diff
@@ -0,0 +1,367 @@
...
+#[derive(Clone)]
+struct NodeContainer {
+    node: Arc<RwLock<Option<Node>>>,
+}
+
+impl Eq for NodeContainer {}
+
+impl PartialEq for NodeContainer {
+    fn eq(&self, other: &Self) -> bool {
+        self.node
```
Mothblocks, line 26: Can you not derive PartialEq and Eq here?

Cargo.toml
```diff
@@ -46,6 +46,9 @@ toml-dep = { version = "0.5.8", package="toml", optional = true }
 aho-corasick = { version = "0.7.18", optional = true}
 rayon = { version = "1.5", optional = true}
 dbpnoise = { version = "0.1.2", optional = true}
+pathfinding = { version = "3.0.13", optional = true }
+mut_static = { version = "5.0.0", optional = true }
```
Mothblocks, line 50: This looks like a wacky library. What do you need it for? We already have both lazy_static and once_cell which should definitely be enough for your use case.

src/pathfinder.rs
```diff
@@ -0,0 +1,367 @@
...
+use pathfinding::prelude::astar;
+use serde::{Deserialize, Serialize};
+use std::borrow::Borrow;
+use std::hash::Hash;
+use std::num::ParseIntError;
+use std::sync::{Arc, RwLock};
+use std::usize::MAX;
+
+lazy_static! {
+    static ref NODES: MutStatic<Vec<NodeContainer>> = MutStatic::new();
```
Mothblocks, line 13: We should phase out lazy_static for once_cell, you should use that instead (we already have it included).

src/pathfinder.rs
```diff
@@ -0,0 +1,367 @@
...
+            }
+        };
+    });
+
+    nodes.push(NodeContainer::new(new_node));
+
+    Ok("1".to_string())
+}
+
+enum DeleteNodeError {
```
Mothblocks, line 184: Remove all your custom stuff and just use [thiserror](https://docs.rs/thiserror/latest/thiserror/) (which we already include).

src/pathfinder.rs
```diff
@@ -0,0 +1,367 @@
...
+fn add_node(json: &str, nodes: &mut Vec<NodeContainer>) -> Result<String, RegisteringNodesError> {
+    let new_node: Node = serde_json::from_str(json)?;
+
+    // As always, a node unique id should correspond to its index in NODES
+    if new_node.unique_id != nodes.len() {
+        return Err(RegisteringNodesError::NodesNotCorrectlyIndexed);
+    }
+
+    // Make sure every connection we have we other nodes is 2 ways
+    new_node.connected_nodes_id.iter().for_each(|index| {
```
Mothblocks, line 171: Just use a normal for loop.

src/pathfinder.rs
```diff
@@ -0,0 +1,367 @@
...
+
+        let node_to_delete_ref = node_to_delete_ref.as_ref().read().unwrap();
+
+        let node_to_delete = match node_to_delete_ref.as_ref() {
+            None => return Err(DeleteNodeError::NodeNotFound),
+            Some(node) => node,
+        };
+
+        // Erase all links to the removed node
+        node_to_delete.connected_nodes_id.iter().for_each(|i| {
```
Mothblocks, line 242: Just use a normal for loop.

src/pathfinder.rs
```diff
@@ -0,0 +1,367 @@
...
+            if let Some(node_container) = nodes.get_mut(*i) {
+                if let Some(node) = node_container.node.write().unwrap().as_mut() {
+                    node.connected_nodes_id
+                        .retain(|index| index != &node_to_delete.unique_id);
+                }
+            }
+        });
+    } // We need to drop everything before set the removed node to None. This is to ensure memory safety
```
Mothblocks, line 250: Woah woah woah, *memory safety*? None of this code is unsafe, what are we worrying about?

src/pathfinder.rs
```diff
@@ -0,0 +1,367 @@
...
+
+    nodes[unique_id] = NodeContainer::new_empty();
+
+    Ok("1".to_string())
+}
+
+enum AstarError {
```
Mothblocks, line 257: Just use thiserror

src/pathfinder.rs
```diff
@@ -0,0 +1,367 @@
...
+byond_fn!(fn generate_path_astar(start_node_id, goal_node_id) {
+    let nodes = NODES.read().unwrap();
+    if let (Ok(start_node_id), Ok(goal_node_id)) = (start_node_id.parse::<usize>(), goal_node_id.parse::<usize>()) {
+        match generate_path(start_node_id, goal_node_id, &nodes) {
+            Ok(vector) => Some(match serde_json::to_string(&vector) {
+                Ok(s) => s,
+                Err(_) => "Cannot serialize path".to_string(),
+            }),
+            Err(e) => Some(match e {
+                AstarError::MutexError(e) => format!("Mutex error : {}", e),
```
Mothblocks, line 279: These should all just be a normal display function (which thiserror makes very easy).

Mothblocks:
A lot of this code scares me.

src/pathfinder.rs
```diff
@@ -0,0 +1,283 @@
...
+        Ok(s) => Some(s),
+        Err(e) => Some(format!("{e}"))
+    }
+});
+
+// Replace the node with unique_id by None
+// Update connected nodes as well so nothing target the removed node anymore
+// Errors if no node can be found with unique_id
+fn remove_node(unique_id: &str) -> Result<String, DeleteNodeError> {
+    let unique_id = match unique_id.parse::<usize>() {
```
BraveMole, line 151: I failed to use ? (error propagation) with the thiserror crate. Did I miss anything or is it impossible ?

src/pathfinder.rs
```diff
@@ -0,0 +1,283 @@
...
+    };
+
+    node_to_delete.connected_nodes_id.iter().for_each(|i| {
+        NODES.with(|nodes_ref| {
+            if let Some(Some(node)) = nodes_ref.borrow_mut().get_mut(*i) {
+                Rc::get_mut(node)
+                    .unwrap()
+                    .connected_nodes_id
+                    .retain(|index| index != &node_to_delete.unique_id);
+            }
```
BraveMole, line 168: Code smell, but I don't really know if it's possible to do better

src/pathfinder.rs
```diff
@@ -0,0 +1,283 @@
...
+use pathfinding::prelude::astar;
+use serde::{Deserialize, Serialize};
+use std::cell::RefCell;
+use std::hash::Hash;
+use std::num::ParseIntError;
+use std::rc::Rc;
+use thiserror::Error;
+
+thread_local! {
+static NODES: RefCell<Vec<Option<Rc<Node>>>> = RefCell::new(Vec::new());
```
BraveMole, line 11: Auxmos maint told me to use thread local, which I admit worked pretty well. 

However he told me that I could also do without the Rc since there's already a RefCell. I do not see how it is possible, but I may be missing something

PR #115, opened 2022-09-04
Set rust-analyzer to all features

BraveMole:
Highly confusing for beginners to not have rust analyzer and not knowing why.


ZeWaka:
my vote is no - this makes it so every time you run a build, rust analyzer will rebuild completely on top of it, leading to both builds taking forever and basic RA functions taking forever, which is *even more* frustrating

if you're not using the default features and want analyzer support, you should add them yourself in your local config - perhaps this could be said explicitly in the readme

ZeWaka:
Closing in favor of 99acb3d, which added a note to the README.

PR #120, opened 2022-10-24
Global panic hook + catch_unwind

### LINKED ISSUE #118: Panicing inside FFI is undefined behavior (and we have plenty of branches that panic)
Need to connect a panic hook or use `dont_panic` (or whatever its called) or something.

Mothblocks:
Fixes #118.

Will log to rustg_panic.log, though I'm not sure where that's going to end up being.


Mothblocks:
Drafted so I can make output path configurable

AffectedArc07:
Dumb question, this is entirely "safe" and "rust native"? 

IE - its not auxtools levels of process hooking or similar, its just normal stuff?

Mothblocks:
I forgot to finish this oops

Yes this is completely fine, `ctor` is a tiny hack but extremely battle tested and absolutely nowhere close to anything like auxtools

Mothblocks:
I'm not sure if this should be failing "silently" like it is now since that is going to cause problems, but that is how it is implemented right now 😓 

Going to update this so that we can get a better grasp of errors on tg

itsmeow:
@Mothblocks My suggestion is just make it up to the individual modules how to handle a panic unwind. My module currently returns the panic string directly, since it normally outputs JSON that is quite easy to test for.

Obviously for something like file_read it's a bit more difficult, but for now I think it's better than crashing the entire game.

ZeWaka:
Mostly implemented in #160, this is pretty abandoned these days.

PR #121, opened 2022-10-25
[Needs testing] 515 Support

AffectedArc07:
Needs testing. I did this entire thing over SSH without access to a BYOND running desktop, but it should work.


dmsrc/main.dm
```diff
@@ -38,5 +38,12 @@
 #define RUST_G (__rust_g || __detect_rust_g())
 #endif
 
+// Handle 515 call() -> call_ext() changes
+#if DM_VERSION >= 515
+#define RGCALL call_ext
+#else
+#define RGCALL call
+#endif
+
```
AffectedArc07, line 47: I dont quite think DM lets me abuse things this much. Does it have an equivalent to Java's `String... args`?

MrStonedOne:
```suggestion
// Handle 515 call() -> call_ext() changes
#if DM_VERSION >= 515
#define RGCALL(RGARGS...) call_ext(RGARGS)
#else
#define RGCALL(RGARGS...) call(...)
#endif


```


> ## Variadic macros
> The last parameter of a macro can end in ... which means that it and all other arguments following it count as a single argument. This is called a variadic macro because it lets you use a variable number of arguments. The last parameter will also become optional.
> 
> ### Example:
> ```dm
> #define DEFAULT_LIST(n, items...) if(!n) n = list(items)
> ```
https://secure.byond.com/docs/ref/#/DM/preprocessor/define

AffectedArc07:
So turns out the preproc needs you to not undefine things used in prior macros, which kinda makes sense.

Commit tree has been cleaned to keep it sane.

PR #122, opened 2022-10-28
add a unix_timestamp extern

ZephyrTFA:
this is for use in https://github.com/tgstation/tgstation/pull/70830

### DIFF
--- a/src/time.rs
+++ b/src/time.rs
@@ -35,3 +35,7 @@ byond_fn!(fn time_reset(instant_id) {
         Some("")
     })
 });
+
+byond_fn!(fn unix_timestamp() {
+    Some(std::time::SystemTime::now().duration_since(std::time::UNIX_EPOCH).unwrap().as_secs().to_string())
+});
--- a/src/time.rs
+++ b/src/time.rs
@@ -36,6 +36,14 @@ byond_fn!(fn time_reset(instant_id) {
     })
 });
 
-byond_fn!(fn unix_timestamp() {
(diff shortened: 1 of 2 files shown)


src/time.rs
```diff
@@ -35,3 +35,7 @@ byond_fn!(fn time_reset(instant_id) {
         Some("")
     })
 });
+
+byond_fn!(fn unix_timestamp() {
+    Some(std::time::SystemTime::now().duration_since(std::time::UNIX_EPOCH).unwrap().as_secs().to_string())
```
Mothblocks, line 40:
```suggestion
    Some(std::time::SystemTime::now().duration_since(std::time::UNIX_EPOCH).unwrap().as_secs_f64().to_string())
```

We would want as much precision as possible for logs

ZeWaka (approved):
will [need](https://doc.rust-lang.org/cargo/reference/semver.html#minor-adding-new-public-items) a minor version bump

Mothblocks:
We're gonna have a few PRs in the next batch so I'll make sure

Mothblocks:
@ZephyrTFA You need to run cargo fmt

ZephyrTFA:
> @ZephyrTFA You need to run cargo fmt

Not currently able to as I am at work, can I convince you to do it for me?

Mothblocks:
I can't I'm at work too

ZephyrTFA:
@Mothblocks formatted

PR #127, opened 2023-02-11
Switch over from libgit2 to gix for git.rs

ZeWaka:
Removes the dependency on libgit2, switches to gitoxide/gix/git-repository instead.

`rustg_git_revparse` working identically, for `rustg_git_revparse("HEAD")` of my local:
image
`rustg_git_commit_date` on that same commit:
image

**Advisory:** This does add a lot of small internal dependencies to the git-repository crate, so the final release binary goes from `6.55MB` to `7.12MB`, an increase of `0.57MB`.
I aim to make up this difference elsewhere (see #128, saved 1.56MB there), as I don't see any immediate ways to pare down this crate (the discover fn is in the base crate, the rest seem to be in deps we could use).

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -18,7 +18,7 @@ byond_fn!(fn rg_git_commit_date(rev) {
         let object = repo.rev_parse_single(rev).map_err(|e| e.to_string())?;
         let commit = object.object().map_err(|e| e.to_string())?.into_commit();
         let time = commit.time().map_err(|e| e.to_string())?;
-        let datetime = Utc.timestamp_opt(time.seconds().into(), 0).earliest().unwrap();
+        let datetime = Utc.timestamp_opt(time.seconds().into(), 0).latest().unwrap();
         Ok(datetime.format("%F").to_string())
     }).ok()
 });
--- a/src/git.rs
+++ b/src/git.rs
@@ -1,17 +1,29 @@
 use chrono::{TimeZone, Utc};
 use git2::{Error, ErrorCode, Repository};
+use git_repository as git;
 
 thread_local! {
+    static REPO: Result<git::Repository, git::discover::Error> = git::discover(".");
     static REPOSITORY: Result<Repository, Error> = Repository::open(".");
 }
 
 byond_fn!(fn rg_git_revparse(rev) {
+    REPO.with(|rep| -> Result<String, String> {
+        let repo = rep.as_ref().map_err(|e| e.to_string())?;
+        let object = repo.rev_parse_single(rev).map_err(|e| e.to_string())?;
+        Ok(object.to_hex().to_string())
+    }).ok()
+});
+
+byond_fn!(fn rg_git_revparseold(rev) {
     REPOSITORY.with(|repo| -> Result<String, ErrorCode> {
         let repo = repo.as_ref().map_err(Error::code)?;
         let object = repo.revparse_single(rev).map_err(|e| e.code())?;
         Ok(object.id().to_string())
     }).ok()
 });
+
+// rustg_git_revparse("origin/master")
 
 byond_fn!(fn rg_git_commit_date(rev) {
     REPOSITORY.with(|repo| -> Result<String, ErrorCode> {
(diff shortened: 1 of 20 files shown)


Mothblocks:
Every time you say libshit2 I have literlaly no idea what you mean and have to check

src/git.rs
```diff
@@ -1,24 +1,25 @@
...
-    static REPOSITORY: Result<Repository, Error> = Repository::open(".");
+    static REPOSITORY: Result<git::Repository, git::discover::Error> = git::discover(".");
 }
 
 byond_fn!(fn rg_git_revparse(rev) {
-    REPOSITORY.with(|repo| -> Result<String, ErrorCode> {
-        let repo = repo.as_ref().map_err(Error::code)?;
-        let object = repo.revparse_single(rev).map_err(|e| e.code())?;
-        Ok(object.id().to_string())
+    REPOSITORY.with(|rep| -> Result<String, String> {
```
Mothblocks, line 9: Why did you make this variable less readable?

src/log.rs
```diff
@@ -31,12 +31,12 @@ byond_fn!(fn log_write(path, data, ...rest) {
             // write first line, timestamped
             let mut iter = data.split('\n');
             if let Some(line) = iter.next() {
-                write!(file, "[{}] {}\n", Utc::now().format("%F %T%.3f"), line)?;
+                writeln!(file, "[{}] {}", Utc::now().format("%F %T%.3f"), line)?;
```
AffectedArc07, line 34: Im like 99% sure its this way to ensure a CRLF ending on windows and linux writes, but I could be completely misunderstanding something.

ZeWaka: Interesting, clippy brought it up so we'd need to mark it ignored.

src/git.rs
ZeWaka: I thought it was clearer that this was the argument, because there is a different local variable (not a shadowed one or anything) named `repo` within the actual block. I can change it back if you want.

src/log.rs
ZeWaka: Nope.
> On all platforms, the newline is the LINE FEED character (`\n`/`U+000A`) alone (no additional CARRIAGE RETURN (`\r`/`U+000D`).
https://doc.rust-lang.org/std/macro.writeln.html

This makes sense, since I don't think clippy would've done that.

AffectedArc07: > Edit: Also - I hope you mean ensuring LF endings.

No we have all logs in CRLF so admins can open them in notepad without the ending being fucked. The clippy thing is valid though.

Mothblocks: This never wrote CRLF

Mothblocks: WindowsTerminal_2023-02-11T18-36-32

src/git.rs
actioninja: Shadowing names in let redefs like this is pretty common in rust code

optimumtact:
>, so the final release binary goes from 7.14MB to 7.55MB, an increase of 0.41MB.

literally who cares lmao I have 4 terabytes of disk space

ZeWaka:
> > , so the final release binary goes from 7.14MB to 7.55MB, an increase of 0.41MB.
> 
> oranges: literally who cares lmao I have 4 terabytes of disk space

see: 
> [6:58 PM] mothblocks: make sure it doesnt bloat our file size too much
[6:58 PM] mothblocks: remembert that loading a dll contributes to the memory limit
[6:58 PM] mothblocks: which isnt as big of a deal as it used to be
[6:58 PM] mothblocks: but im sure someone willbe p issed off at me eventually about it

Mothblocks:
That was in reference to statically linking an ssl lib btw

optimumtact:
who would be pissed what a load of horse shit

Mothblocks:
Idk i just dont like getting yelled at by mso

optimumtact:
protip fo rdealing with mso just lie and say I did it

ZeWaka:
need to switch to gix package name

optimumtact:
finish ur shit

ZeWaka:
This isn't a super big problem anymore since libgit2 doesn't require libssl1 (`mysql` still does unfortunately) anymore it seems. This PR could go either way.

optimumtact:
replace mysql with postgresql

ZeWaka:
Closing due to redundant

PR #131, opened 2023-05-13
Replace flume with crossbeam in HTTP

Mothblocks:
We're getting a really weird panic in this area of the code and it seems to be a result of flume unwinding.

This is a shot in the dark experiment to see if using the much more popular and battle-tested crossbeam will resolve this issue. Long term we need to agree on what panic hooks should do (exit DD? try to handle when an error is expected (like HTTP)? just return *something* and let DM code fail silently?) and get them in. If this PR actually should be merged because flume is erroneous, then we need to remove it completely in redis as well, so this is a draft.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -20,13 +20,6 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2
 
-      - name: Clippy (all features)
-        uses: actions-rs/cargo@v1
-        with:
-          toolchain: stable
-          command: clippy
-          args: --target i686-pc-windows-msvc --all-features --locked -- -D warnings
-
       - name: Rustfmt
         uses: actions-rs/cargo@v1
         with:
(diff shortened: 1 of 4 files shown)


AffectedArc07:
Dumb/unrelated question but would this solve the issue we have where we need dedicated `start_http_client()` and `shutdown_http_client()` methods to wipe out old http clients after a round between DD restarts? We had some issues with safety and bad memory accesses until we did this approach.

Relevant patch: https://github.com/ParadiseSS13/rust-g/blob/master/paradise-rust-g-patches/0005-Paradise-HTTP-Module-Tweaks.patch

Bizzonium:
Also please check out these 3 commits, it could give some inspiration.
https://github.com/ss220-space/rust-g-tg/compare/a4800676b265337ff4ade5534c6692c678e7579a...fcc86bc4b0432d7d41bf86c3309e119b3c01db81

Mothblocks:
This didn't actually fix it and the stack traces we're getting are the same. I'm going to get my panic hooks PR in better shape and get a better understanding

PR #132, opened 2023-05-14
New FFI Macro

actioninja:
Brand new ludicrously overengineered macro for automagically™ interoping with byond smoothly and cleanly.

## Here's what you can do with the new macro
All of these are callable directly from byond.
### Write bare functions without being wrapped in a macro
```rust
#[byond_fn]
pub fn example_byond_fn() {
    let x = "yeah, we can bind vars";
    println!("{x}");
}
```
### Have string arguments and return strings
```rust
#[byond_fn]
pub fn concat(first: &str, second: &str) -> String {
    format!("{first}{second")
}
```
### Have arbitrary args that can be parsed from strings and return arbitrary types that can be converted to strings
```rust
#[byond_fn]
pub fn add(arg1: u8, arg2: u8) -> u8 {
    arg1 + arg2
}
```
### Have some arguments be optional and not required to be passed
```rust
#[byond_fn]
pub fn example_optional_params(arg1: u8, arg2: Option<u8>) -> u8 {
    arg1 + arg2.unwrap_or(0)
}
```

### Gracefully handle errors
```rust
#[byond_fn]
pub fn do_something_falliable(dunno: &str) -> Result<String, WhateverError> {
    let cool = probably_something_cool(dunno)?;
    let oh_no = actually_it_was_lame(cool)?;
    Ok(oh_no)
}
```

### Automagic JSON Transport
```rust
pub fn fungle_the_goobler(goobler: Json<Goobler>) -> Json<Fungled> {
    let fungled = goobler.into_inner().fungle();
    Json(fungled)
}
```

New macro lives here (at the moment): https://github.com/actioninja/byond_fn

Docs here: https://docs.rs/byond_fn/0.5.1/byond_fn/

## How the hell does this work?

In a very overcomplicated way

The tl;dr is it uses a trait to parse arguments from strings, and a separate trait to serialize back to string. There's blanket impls over Option and Result as well for optional args and error handling respectively.

This will be better documented later, read the code, the docs, etc

## Error handling

This is very much in flux right now. Especially considering that the best way to handle this may change when FFI V2 becomes available. How it currently works can be found in the docs (though these docs may not be accurate atm): https://docs.rs/byond_fn/0.5.1/byond_fn/str_ffi/index.html#errors

## Goals

I never really was super happy with the existing macro. Since it puts the function body inside a macro call, you lose a lot of IDE niceness since inside macros could be anything. On top of that, it always was kind of weird to me that each function was expected to just do the boilerplate of parsing each string argument manually when this is going to be the same in the overwhelming amount of usecases.

The unclear and not well documented nature of the previous macro resulted in a lot of cargo cult as well. Lots of functions that return results despite having no fallibility.

Someone mentioned in coderbus a while ago that it would be nice if the actual FFI junk could be pulled out of rust-g so it could be reused elsewhere more easily. Since proc macros require their own crate anyways, this makes a good opportunity to do so.

## `byond_fn` todo

- [ ] Finalize the argument structure and how FFI types should be specified
- [ ] Finalize the error handling structure
- [ ] Implement some way of automatically generating the byond-side call signature (companion crate for usage in build.rs?)
- [ ] Fully cover all pub api with documentation
- [ ] Write some basic unit tests
- [ ] Implement the new byond FFI interface

# Things that should happen before this is merged

- [ ] finalize `byond_fn` structure and output and hit 1.0 (above todo list should be finished)
- [ ] convert all existing functions in rust-g to use the new macro
- [ ] bring `byond_fn` under the tgstation org and give more accounts crates.io publish perms
- [ ] run a built binary using the new macro in prod for a while to check against regressions

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -60,7 +60,7 @@ checksum = "96cf8829f67d2eab0b2dfa42c5d0ef737e0724e4a82b01b3e292456202b19716"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 1.0.98",
 ]
 
 [[package]]
@@ -149,6 +149,25 @@ name = "bumpalo"
 version = "3.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "37ccbd214614c6783386c1af30caf03192f17891059cecc394b4fb119e363de3"
+
+[[package]]
+name = "byond_fn"
+version = "0.4.0"
+dependencies = [
+ "byond_fn_impl",
+ "serde",
+ "serde_json",
+]
+
+[[package]]
+name = "byond_fn_impl"
+version = "0.4.0"
+dependencies = [
+ "proc-macro-error",
+ "proc-macro2",
+ "quote",
+ "syn 2.0.15",
+]
 
 [[package]]
 name = "bytemuck"
@@ -460,7 +479,7 @@ checksum = "532b4c15dccee12c7044f1fcad956e98410860b22231e44a3b827464797ca7bf"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 1.0.98",
 ]
 
 [[package]]
@@ -590,7 +609,7 @@ checksum = "3dbc4f084ec5a3f031d24ccedeb87ab2c3189a2f33b8d070889073837d5ea09e"
 dependencies = [
  "frunk_proc_macro_helpers",
  "quote",
- "syn",
+ "syn 1.0.98",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


AffectedArc07:
`fungle_the_goobler`

Cyprex:
How (un)performant is this?

actioninja:
> How (un)performant is this?

Shouldn't be meaningfully different from the existing macro. The two potential vectors for performance differences are the extra error handling and sanity checking (of which this mostly consists of a check to make sure the amount of arguments is correct), and the extra trait function calls which are all static dispatch anyways.

AffectedArc07:
FFIV2 is out - Tiger is currently making a crate for it https://github.com/spacestation13/byondapi-rs

PR #133, opened 2023-05-19
Adds a redis reliable queue implementation

### LINKED ISSUE #129: Add Redis list operation support
Pub/sub is great for events if you have something listening, but not so great if you care about a historical record of the data if no one happens to be subscribed at the time. For such a use case, lists are better suited combined with a worker-style implementation via BRPOP.

Would be neat to have the ability to push to, and pop off, lists.

(LPUSH and LRANGE)

ZeWaka:
fixes #129

Instead of just having a redis pub/sub model, this lets us (goonstation at least) use redis as a reliable event queue. Basically, we want to shit tons of our event logging and stuff to redis to eventually be consumed by our api and fancy~ management bits.

image

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,6 @@
 {
-    "rust-analyzer.cargo.target": "i686-pc-windows-msvc"
+    "rust-analyzer.cargo.target": "i686-pc-windows-msvc",
+    "rust-analyzer.linkedProjects": [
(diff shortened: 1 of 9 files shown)


AffectedArc07:
This is giving me ideas of cool things to do.

I could refactor the entire queue system with this, hmmmmmmmmm.

optimumtact:
this is drafted, why?

PR #137, opened 2023-06-02
Some pedantic and nursery clippy lints

ZeWaka:
Does a bunch of clippy pedantic and nursery lints, mostly just cleanliness/idiomatic stuff. No functionality changes (other than things like ""performance"" due to using borrowed instead of copied values).

Should this be merged?
image

### DIFF
--- a/src/cellularnoise.rs
+++ b/src/cellularnoise.rs
@@ -71,10 +71,10 @@ fn noise_gen(
     });
 
     //then we cut it
-    let map = (1..width + 1)
+    let map = (1..=width)
         .into_par_iter()
         .map(|x| {
-            (1..height + 1)
+            (1..=height)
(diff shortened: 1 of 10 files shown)


src/cellularnoise.rs
```diff
@@ -71,10 +71,10 @@ fn noise_gen(
     });
 
     //then we cut it
-    let map = (1..width + 1)
+    let map = (1..=width)
         .into_par_iter()
         .map(|x| {
-            (1..height + 1)
+            (1..=height)
```
MrStonedOne, line 77: sometimes its hard to take rust seriously.

actioninja: This operator used to be `...` instead of `..=`

PR #140, opened 2023-06-23
adds rustg_dmi_icon_states

silicons:
coderbus message link https://canary.discord.com/channels/484170914754330625/484170915253321734/1121523986996666519

tl;dr: faced with needing to wait for byond to un-heck this, and the fact that citadel is burning 400MB on icon_states right now, welcome to baby's first ""production"" rust PR

all this adds is rustg_dmi_icon_states which will hopefully be better than byond's icon_states for a time to come. all it does is open the dmi path, extract states, return a JSON

in theory this contains no memory leaks/weirdness since rust should handle the file closes/drops but uh please scrutinize this PR carefully i cannot understate how new I am to rust.

this also adds rohesie's `dmi` as that's what handles the uh, actual dmi loading.

### DIFF
--- a/dmsrc/dmi.dm
+++ b/dmsrc/dmi.dm
@@ -2,4 +2,5 @@
 #define rustg_dmi_create_png(path, width, height, data) RUSTG_CALL(RUST_G, "dmi_create_png")(path, width, height, data)
 #define rustg_dmi_resize_png(path, width, height, resizetype) RUSTG_CALL(RUST_G, "dmi_resize_png")(path, width, height, resizetype)
 /// output: json_encode'd list. json_decode to get a flat list with icon states in the order they're in inside the .dmi
+/// input must be a path, not an /icon; you have to do your own handling if it is one, as icon objects can't be directly passed to rustg.
 #define rustg_dmi_icon_states(fname) RUSTG_CALL(RUST_G, "dmi_icon_states")(fname)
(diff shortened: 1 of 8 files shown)


dmsrc/dmi.dm
```diff
@@ -1,3 +1,6 @@
 #define rustg_dmi_strip_metadata(fname) RUSTG_CALL(RUST_G, "dmi_strip_metadata")(fname)
 #define rustg_dmi_create_png(path, width, height, data) RUSTG_CALL(RUST_G, "dmi_create_png")(path, width, height, data)
 #define rustg_dmi_resize_png(path, width, height, resizetype) RUSTG_CALL(RUST_G, "dmi_resize_png")(path, width, height, resizetype)
+/// output: json_encode'd list. json_decode to get a flat list with icon states in the order they're in inside the .dmi
+/// input must be a path, not an /icon; you have to do your own handling if it is one, as icon objects can't be directly passed to rustg.
```
ZeWaka, line 5: Should use the multiline doc comment syntax, like https://github.com/tgstation/rust-g/blob/master/dmsrc/cellularnoise.dm

src/dmi.rs
```diff
@@ -112,3 +118,20 @@ fn resize_png<P: AsRef<Path>>(
 
     Ok(newimg.save_with_format(path.as_ref(), image::ImageFormat::Png)?)
 }
+
+/// output is a JSON string for reading within BYOND
+/// erroring at any point will produce an empty string
```
ZeWaka, line 123: Why not produce an actual error instead of an empty string?

ZeWaka (changes requested):

dmsrc/dmi.dm
ZeWaka: Also, it's not my dmi lib, rohesie made it

ZeWaka: You'll also want to switch to the new `0.3.0` version

ZeWaka (approved):

PR #144, opened 2023-07-17
Make mysql use the native rust tls stack by default

optimumtact:
This removes the dependency on libssl, which is one of the few remaining dependencies that it has beyond libc

This is a major version bump because there may be some inconsistencies with the tls certs suported by rustls

https://docs.rs/mysql/latest/mysql/#ssl-support

You can opt back into the native stack if you so choose

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "rust-g"
 edition = "2021"
-version = "3.0.0"
+version = "2.1.0"
 authors = [
     "Bjorn Neergaard <bjorn@neersighted.com>",
     "Tad Hardesty <tad@platymuus.com>",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "rust-g"
 edition = "2021"
-version = "2.1.0"
+version = "3.0.0"
 authors = [
     "Bjorn Neergaard <bjorn@neersighted.com>",
     "Tad Hardesty <tad@platymuus.com>",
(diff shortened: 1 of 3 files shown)


optimumtact:
can anyone explain to me why cargo is ignoring the updated cargo lock file?

ZeWaka:
heh I had a branch for this but got busy, will review soon

AffectedArc07:
This is hot

Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "rust-g"
 edition = "2021"
-version = "2.1.0"
+version = "3.0.0"
```
ZeWaka, line 4: normally done for the release tag PR version (ex. https://github.com/tgstation/rust-g/pull/143), but can do here.

ZeWaka (approved):

Cargo.toml
```diff
@@ -46,7 +46,7 @@ serde = { version = "1.0", optional = true, features = ["derive"] }
 serde_json = { version = "1.0", optional = true }
 lazy_static = { version = "1.4", optional = true }
 once_cell = { version = "1.17", optional = true }
-mysql = { version = "24.0", optional = true }
+mysql = { version = "24.0", default_features = false, optional = true}
```
ZeWaka, line 49: just making sure - we don't need the `rustls-tls` right?

optimumtact: we set the featureset used for mysql with a feature of our own named rustls-tls that is in the default set, so the import specifies no features at all.

ZeWaka: yep, just making sure

PR #148, opened 2023-08-02
Revert "3.0.0"

MrsTonedOne:
Reverts tgstation/rust-g#147

Makes the server crash on round restart on windows.

This kills pop counts because people don't always reconnect and its common for players to decide to end a play session if they meet the slightest resistance.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -63,7 +63,7 @@ jobs:
       - run: |
           sudo dpkg --add-architecture i386
           sudo apt-get update
-          sudo apt-get install libgcc-s1:i386 g++-multilib zlib1g-dev:i386
+          sudo apt-get install libgcc-s1:i386 g++-multilib zlib1g-dev:i386 libssl-dev:i386
           ./scripts/install_byond.sh
 
       - uses: actions-rs/toolchain@v1
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2166,7 +2166,7 @@ dependencies = [
 
 [[package]]
 name = "rust-g"
-version = "3.0.0"
+version = "2.1.0"
 dependencies = [
  "aho-corasick",
  "base64",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "rust-g"
 edition = "2021"
-version = "3.0.0"
+version = "2.1.0"
 authors = [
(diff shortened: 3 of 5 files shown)


optimumtact:
do you have a stack trace?

MrStonedOne:
No, stack traces go to stdout/stderr which only tgsv5 will save to disk and the only tgsv5 server is on linux which doesn't have this issue.

optimumtact:
Then how do you know it's related to rust-g?

optimumtact:
i mean the stack trace from the minidump

MrStonedOne:
the timing matches up with the first 3.0.0 test merges and the fault offset shows up as rustg

optimumtact:
they were crashing on the 25th and earlier too though, in fact you can see crashes on reboots all through july.

MrStonedOne:
yes, randomly, not consistently. the crash rate goes from every few rounds to every round for sybil on the 18th, which is when the first test merge happened on sybil.

optimumtact:
If I had to guess it's something to do with native mysql rust-tls probably using more tokio and causing crashes on exit.

ZeWaka (commented):
reverting a released version? lmao no just downgrade on tg
i can yank the release if you really are scared of people downloading it, but I encountered no real problems testing this locally and it's probably just the generic 'all crashes are blamed on rust-g due to open connections'

MrStonedOne:
ya if you bothered to read the 

> it's probably just the generic 'all crashes are blamed on rust-g due to open connections'

ya if you bothered to read the pr comments you'd know that its being blamed because the timing matches up.

also my concerns when i open revert fix prs are always generally about fixing master, i can't pr to the release tab sorry.

ZeWaka:
> ya if you bothered to read the pr comments you'd know that its being blamed because the timing matches up.

what comments, how am i supposed to know? tgstation/tgstation/pull/77307 has nothing and tgstation/tgstation/pull/76663 has nothing, and neither are linked here

optimumtact:
this isn't how you fix versioned packages.

The right answer would be a 3.0.1 which changes the cargo build script to use nativetls again.

But the problem is there's no definitive cause for the issue and i can't get enough information from the dmp file to actually determine what the root cause is.

MrStonedOne:
byond has fixed the reconnect topic bug so crashing during restart is no longer a concern for pop counts.

also this might be fixed.

PR #149, opened 2023-08-09
Add dmi2svg as a function that returns css in a string

tigercat2000:
I'm using this to make `/datum/asset/spritesheet_svg`, see [this](https://github.com/ShadowLarkens/CHOMPStation2/commit/7917fbd23d824dad21f86af5745e0d29b40869a6#diff-9b53cb11218b0be4234bad019e2c8816727577284da071c847c01c34c2fa7b32R114) for an example of usage.

tl;dr
```dm
fcopy(icon_you_want_svgs_of, "icon.dmi");
// first argument is source file, second argument is target css
// third is an associative map of names.
// if you want to, for example, change "icon-state-a" to "statea", you would do `json_encode(list("icon-state-a" = "statea"))`
rustg_dmi_convert_to_svg("icon.dmi", "icon.css", json_encode(list()))
// If you want the actual CSS text in memory:
var/css = file2text("icon.css")
// Otherwise if you're just sending it to a client:
fcopy_rsc("icon.css")
```

Basically, this uses my new library, [`dmi2svg`](https://github.com/tigercat2000/dmi2svg), to take a dmi file and convert all of the sprites within into svg paths using the `contour_tracing` library.

Known limitations right now:
 - Animated icons are not supported
 - Directional icons have to be split into separate states if you want to render each direction.

If you provide a DMI with either of these, it just grabs the first frame facing SOUTH and discards the rest. This is already good enough for my usecase, but I'm open to supporting these if people would like them.

Cargo.lock got updated because of a conflict in `thiserror` version requirements.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,6 +57,7 @@ dbpnoise = { version = "0.1.2", optional = true }
 pathfinding = { version = "4.2.1", optional = true }
 num = { version = "0.4.0", optional = true }
 dmi = { version = "0.3.0", optional = true }
+dmi2svg = { git = "https://github.com/tigercat2000/dmi2svg", optional = true }
 
 [features]
 default = [
@@ -103,7 +104,7 @@ all = [
 # default features
 acreplace = ["aho-corasick"]
 cellularnoise = ["rand", "rayon"]
-dmi = ["png", "image", "dep:dmi"]
+dmi = ["png", "image", "dep:dmi", "dep:dmi2svg", "base64"]
 file = []
 git = ["git2", "chrono"]
 http = ["reqwest", "serde", "serde_json", "once_cell", "jobs"]
--- a/dmsrc/dmi.dm
+++ b/dmsrc/dmi.dm
@@ -7,3 +7,12 @@
  * output: json_encode'd list. json_decode to get a flat list with icon states in the order they're in inside the .dmi
  */
 #define rustg_dmi_icon_states(fname) RUSTG_CALL(RUST_G, "dmi_icon_states")(fname)
+/**
+ * input:
+ * fname: must be a path, not an /icon; you have to do your own handling if it is one, as icon objects can't be directly passed to rustg.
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -57,6 +57,7 @@ dbpnoise = { version = "0.1.2", optional = true }
 pathfinding = { version = "4.2.1", optional = true }
 num = { version = "0.4.0", optional = true }
 dmi = { version = "0.3.0", optional = true }
+dmi2svg = { git = "https://github.com/tigercat2000/dmi2svg", optional = true }
```
ZeWaka, line 60: this should be [set to a release](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories) using `rev` or `tag`

dmsrc/dmi.dm
```diff
@@ -7,3 +7,33 @@
...
+ * MUST CALL rustg_dmi_start_svg_symbols and rustg_dmi_end_svg_symbols
+ * 
+ * input:
+ * fname: must be a path, not an /icon; you have to do your own handling if it is one, as icon objects can't be directly passed to rustg.
+ * output: path to output the svg <symbol> list to
+ * namemap: must be a json_encode'd list of {'icon_state': 'desired name'} or ""
+ *
+ * output: ready-to-ship svg file
+ */
+#define rustg_dmi_convert_to_svg_symbols(fname, output, namemap) RUSTG_CALL(RUST_G, "dmi_convert_to_svg_symbols")(fname, output, namemap)
```
ZeWaka, line 34: what's with the _symbols suffix?
Bit unclear what's the actual difference between the 'svg symbols' one and the 'svg css' one - why use either

ZeWaka (changes requested):

tigercat2000:
This seems unnecessary now that 516 has Edge and we can just nearest neighbor upscale stuff.

PR #150, opened 2023-09-25
Standardize redis_reliablequeue connect/disconnect output

ZeWaka:
Essentially makes it match the 
```json
{
  "success": bool
  "content": string
}
```
format instead of just returning the raw rust error through `?` propagation.


ZeWaka:
should probably test this

AffectedArc07:
> should probably test this

I can TM this on para, I know we use redis pubsub.

ZeWaka:
i intend to get to this sometime now, since goon uses this in prod

ZeWaka:
x

PR #151, opened 2023-09-25
Restructure redis_reliablequeue to keep the connection open

ZeWaka:
title, basically don't drop the connection and keep it open to hopefully reduce overhead

need to test still and measure timings


ZeWaka:
Closing this since we don't really need it and the overhead is only like 80ms per transaction

PR #154, opened 2023-10-24
Adds Base64 decoding stuff

AffectedArc07:
Title says it all. Tested and works as expected, and thrown a test into the repo.

This was surprisingly easy because we already have the b64 crate for encoding, just needed to add a decode method.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "rust-g"
 edition = "2021"
-version = "3.0.0"
+version = "3.1.0"
 authors = [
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "rust-g"
 edition = "2021"
-version = "3.0.0"
+version = "3.1.0"
```
ZeWaka, line 4: we generally don't update versions in feature PRs

AffectedArc07: Ok

PR #155, opened 2023-10-29
Shutdown implementation which hooks into /world/Del

ZephyrTFA:
Add shutdown handling for rust-g

Move log closing to said proc.
Implement job detaching for said proc.

### DIFF
--- a/dmsrc/log.dm
+++ b/dmsrc/log.dm
@@ -1,2 +1,0 @@
-#define rustg_log_write(fname, text, format) RUSTG_CALL(RUST_G, "log_write")(fname, text, format)
-/proc/rustg_log_close_all() return RUSTG_CALL(RUST_G, "log_close_all")()
--- a/dmsrc/log.dm
+++ b/dmsrc/log.dm
@@ -1,0 +1,1 @@
+#define rustg_log_write(fname, text, format) RUSTG_CALL(RUST_G, "log_write")(fname, text, format)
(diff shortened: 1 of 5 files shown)


dmsrc/log.dm
```diff
@@ -1,2 +0,0 @@
-#define rustg_log_write(fname, text, format) RUSTG_CALL(RUST_G, "log_write")(fname, text, format)
```
ZeWaka, line 1: no

ZeWaka (changes requested):

AffectedArc07:
Wonder if this would fix our need to have a patch on the HTTP client that has a designated create and shutdown method.

dmsrc/world.dm
```diff
@@ -0,0 +1,3 @@
+/world/Del(...)
+	RUSTG_CALL(RUST_G, "shutdown_rustg")
```
ZeWaka, line 2: This'll end up just copying itself  in the rust_g.dm file right? It needs documentation on what it's doing and why it's doing it then.
Some codebases are particular about ordering. See: you don't want this before the `logs << "shutting down"` logging entry.

ZeWaka (changes requested):

optimumtact:
where do they go when they get detatched.

ZephyrTFA:
> where do they go when they get detatched.

No where, we just drop our reference to them which means that they become free standing threads. There's no easy way to hook cancelation logic in and leaving them orphaned won't cause any problems

willox:
"Detaching" the threads would mean they may still exist after the module is unloaded which will result in a crash. It doesn't do anything.

PR #158, opened 2023-12-20
Updates `rustg_file_exists()` macro to remove user error

### LINKED ISSUE #153: rustg_file_exists() returns "false" instead of "" if the file is not found
Rustg methods that replace byond methods should behave identically or only expand on the behavior of the byond method. A github code search of "rustg_file_exists" shows that every single user of this method thinks it returns a truthy or falsey value. However, this is not the case.

Kapu1178:
Closes #153 

Creates truthy-ness parity with byond's fexists() function.


Mothblocks (changes requested):
image

PR #160, opened 2023-12-23
IconForge - Building spritesheets at the speed of light

itsmeow:
Adds IconForge, a new module for building asset spritesheets.

It's over 10x faster than BYOND in some cases. This module just takes a JSON input and spits out a spritesheet to a path, so it's up to the DM side to actually use it correctly, but I've tested it with a branch on Bee: https://github.com/itsmeowForks/BeeStation-Hornet/tree/asset-rustg

It is *quite* fast.
image

I've also added a panic handler to the core rustg module, since this module is.. ~~prone to panicing~~ Not anymore, but it was useful in testing. This should also help diagnose crashes from within rust-g without needing to take minidumps.

~~In the future I want to add a more advanced caching system to this, but as it is~~, it's already just a faster replacement for the existing spritesheet generation code in DM, and people are interested, so here it is. The way the JSON input is designed allows you to precisely determine if anything has changed, since it only supports DMI+icon_states and not any BYOND /icons. You can basically just hash everything and compare to get better caching.

Also adds a caching validation tool. This hashes all DMIs within the `sprites` object and compares them to a previous one provided by DM. It also provides the hashes as the result of the generate() function when prompted. It is up to DM to do the rest, rustg only provides a fast method for comparing the hashes with parallelism.

Cache validation check doesn't cost too much:
image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -146,7 +146,8 @@ iconforge = [
     "tracy_full",
     "jobs",
     "once_cell",
-    "dashmap"
+    "dashmap",
+    "hash",
 ]
 redis_pubsub = ["flume", "redis", "serde", "serde_json"]
 redis_reliablequeue = ["flume", "redis", "serde", "serde_json"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,6 +63,7 @@ dbpnoise = { version = "0.1.2", optional = true }
 pathfinding = { version = "4.4", optional = true }
 num-integer = { version = "0.1.45", optional = true }
 dmi = { version = "0.3.1", optional = true }
+tracy_full =  { version = "1.6.1", optional = true}
 
 [features]
 default = [
@@ -134,6 +135,7 @@ hash = [
     "serde_json",
 ]
 pathfinder = ["num-integer", "pathfinding", "serde", "serde_json"]
+iconforge = ["serde", "serde_json", "png", "image", "dep:dmi", "rayon", "tracy_full", "jobs", "once_cell"]
 redis_pubsub = ["flume", "redis", "serde", "serde_json"]
 redis_reliablequeue = ["flume", "redis", "serde", "serde_json"]
 unzip = ["zip", "jobs"]
@@ -148,3 +150,16 @@ jobs = ["flume"]
 
 [dev-dependencies]
 regex = "1"
+
+# DEV ONLY. REMOVE THIS LATER
+[profile.dev.package.dmi]
+opt-level = 3
+
+[profile.dev.package.image]
+opt-level = 3
+
+[profile.dev.package.serde]
+opt-level = 3
+
+[profile.dev.package.serde_json]
+opt-level = 3
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -91,6 +91,7 @@ all = [
     "file",
     "git",
     "http",
+    "iconforge",
     "json",
     "log",
     "noise",
@@ -135,7 +136,6 @@ hash = [
     "serde",
     "serde_json",
 ]
-pathfinder = ["num-integer", "pathfinding", "serde", "serde_json"]
 iconforge = [
     "serde",
     "serde_json",
@@ -149,6 +149,7 @@ iconforge = [
     "dashmap",
     "hash",
 ]
+pathfinder = ["num-integer", "pathfinding", "serde", "serde_json"]
 redis_pubsub = ["flume", "redis", "serde", "serde_json"]
 redis_reliablequeue = ["flume", "redis", "serde", "serde_json"]
 unzip = ["zip", "jobs"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,6 +64,7 @@ pathfinding = { version = "4.4", optional = true }
 num-integer = { version = "0.1.45", optional = true }
 dmi = { version = "0.3.1", optional = true }
 tracy_full =  { version = "1.6.1", optional = true}
+#tracy_full =  { version = "1.6.1", optional = true, features = ["enable"]}
 
 [features]
 default = [
@@ -135,7 +136,18 @@ hash = [
     "serde_json",
 ]
 pathfinder = ["num-integer", "pathfinding", "serde", "serde_json"]
-iconforge = ["serde", "serde_json", "png", "image", "dep:dmi", "rayon", "tracy_full", "jobs", "once_cell"]
+iconforge = [
+    "serde",
+    "serde_json",
+    "png",
+    "image",
+    "dep:dmi",
+    "rayon",
+    "tracy_full",
+    "jobs",
+    "once_cell",
+    "dashmap"
+]
 redis_pubsub = ["flume", "redis", "serde", "serde_json"]
 redis_reliablequeue = ["flume", "redis", "serde", "serde_json"]
 unzip = ["zip", "jobs"]
@@ -150,16 +162,3 @@ jobs = ["flume"]
 
 [dev-dependencies]
 regex = "1"
-
-# DEV ONLY. REMOVE THIS LATER
-[profile.dev.package.dmi]
-opt-level = 3
-
-[profile.dev.package.image]
-opt-level = 3
-
-[profile.dev.package.serde]
-opt-level = 3
-
-[profile.dev.package.serde_json]
-opt-level = 3
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,7 @@ serde_json = { version = "1.0", optional = true }
 lazy_static = { version = "1.4", optional = true }
 once_cell = { version = "1.19", optional = true }
 mysql = { version = "24.0", default_features = false, optional = true }
-dashmap = { version = "5.5", optional = true }
+dashmap = { version = "5.5", optional = true, features = ["rayon", "serde"]}
 zip = { version = "0.6", optional = true }
 rand = { version = "0.8", optional = true }
 toml-dep = { version = "0.8.8", package = "toml", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,9 +62,9 @@ rayon = { version = "1.8", optional = true }
 dbpnoise = { version = "0.1.2", optional = true }
 pathfinding = { version = "4.4", optional = true }
 num-integer = { version = "0.1.45", optional = true }
-dmi = { version = "0.3.3", optional = true }
-tracy_full = { version = "1.6.1", optional = true }
-#tracy_full = { version = "1.6.1", optional = true, features = ["enable"] }
+dmi = { version = "0.3.4", optional = true }
+#tracy_full = { version = "1.6.1", optional = true }
+tracy_full = { version = "1.6.1", optional = true, features = ["enable"] }
 
 [features]
 default = [
@@ -139,6 +139,7 @@ hash = [
 iconforge = [
     "dashmap",
     "dep:dmi",
+    "hash",
     "image",
     "jobs",
     "once_cell",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,7 +62,7 @@ rayon = { version = "1.8", optional = true }
 dbpnoise = { version = "0.1.2", optional = true }
 pathfinding = { version = "4.4", optional = true }
 num-integer = { version = "0.1.45", optional = true }
-dmi = { version = "0.3.1", optional = true }
+dmi = { version = "0.3.3", optional = true }
 tracy_full =  { version = "1.6.1", optional = true}
 #tracy_full =  { version = "1.6.1", optional = true, features = ["enable"]}
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,8 +63,7 @@ dbpnoise = { version = "0.1.2", optional = true }
 pathfinding = { version = "4.4", optional = true }
 num-integer = { version = "0.1.45", optional = true }
 dmi = { version = "0.3.4", optional = true }
-#tracy_full = { version = "1.6.1", optional = true }
-tracy_full = { version = "1.6.1", optional = true, features = ["enable"] }
+tracy_full = { version = "1.6.1", optional = true }
 
 [features]
 default = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,7 @@ serde_json = { version = "1.0", optional = true }
 lazy_static = { version = "1.4", optional = true }
 once_cell = { version = "1.19", optional = true }
 mysql = { version = "24.0", default_features = false, optional = true }
-dashmap = { version = "5.5", optional = true, features = ["rayon", "serde"]}
+dashmap = { version = "5.5", optional = true, features = ["rayon", "serde"] }
 zip = { version = "0.6", optional = true }
 rand = { version = "0.8", optional = true }
 toml-dep = { version = "0.8.8", package = "toml", optional = true }
@@ -63,8 +63,8 @@ dbpnoise = { version = "0.1.2", optional = true }
 pathfinding = { version = "4.4", optional = true }
 num-integer = { version = "0.1.45", optional = true }
 dmi = { version = "0.3.3", optional = true }
-tracy_full =  { version = "1.6.1", optional = true}
-#tracy_full =  { version = "1.6.1", optional = true, features = ["enable"]}
+tracy_full = { version = "1.6.1", optional = true }
+#tracy_full = { version = "1.6.1", optional = true, features = ["enable"] }
 
 [features]
 default = [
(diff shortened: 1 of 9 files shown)


itsmeow:
Also please be nice, this is my first Rust project ever. I know the code is kinda messy.

src/iconforge.rs
```diff
@@ -0,0 +1,551 @@
...
+        let dmi_arc = Arc::new(dmi.unwrap());
+        let other_arc = dmi_arc.clone();
+        // cache it for later.
+        // Ownership is given to the hashmap
+        ICON_FILES.insert(icon_path.to_owned(), dmi_arc);
+        Ok(other_arc)
+    }
+}
+
+// Takes an IconObject, gets its DMI, then picks out a DynamicImage for the IconState, as well as transforms the DynamicImage.
```
ZeWaka, line 286: Use doc comments

src/iconforge.rs
```diff
@@ -0,0 +1,551 @@
...
+const NORTH: u8 = 1;
+const EAST: u8 = 4;
+const WEST: u8 = 8;
+const FOUR_DIRS: [u8; 4] = [SOUTH, NORTH, EAST, WEST];
+const SOUTHEAST: u8 = SOUTH | EAST; // 6
+const SOUTHWEST: u8 = SOUTH | WEST; // 10
+const NORTHEAST: u8 = NORTH | EAST; // 5
+const NORTHWEST: u8 = NORTH | WEST; // 9
+// This is ordered by how DMIs internally place dirs into the PNG
+const EIGHT_DIRS: [u8; 8] = [SOUTH, NORTH, EAST, WEST, SOUTHEAST, SOUTHWEST, NORTHEAST, NORTHWEST];
```
ZeWaka, line 33: Probably should be in some shared module or it's own lib tbh

src/iconforge.rs
```diff
@@ -0,0 +1,551 @@
...
+            });
+        return Err(Error::IconState(message.unwrap().to_owned()));
+    }
+    x.ok().unwrap()
+}
+
+fn generate_spritesheet(file_path: &str, spritesheet_name: &str, sprites: &str) -> std::result::Result<String, Error> {
+    zone!("generate_spritesheet");
+
+    let error: Arc<Mutex<Vec<String>>> = Arc::new(Mutex::new(Vec::new()));
```
ZeWaka, line 142: For a lot of this stuff, you don't need the explicit types written out when initialized like this

src/iconforge.rs
itsmeow: It should really be part of the DMI crate, heh

itsmeow:
DNM until https://github.com/spacestation13/dmi-rust/pull/20

Cargo.toml
```diff
@@ -53,7 +53,7 @@ serde_json = { version = "1.0", optional = true }
 lazy_static = { version = "1.4", optional = true }
 once_cell = { version = "1.19", optional = true }
 mysql = { version = "24.0", default_features = false, optional = true }
-dashmap = { version = "5.5", optional = true }
+dashmap = { version = "5.5", optional = true, features = ["rayon", "serde"]}
```
AffectedArc07, line 56:
```suggestion
dashmap = { version = "5.5", optional = true, features = ["rayon", "serde"] }
```

itsmeow:
If you need proof of this thing's pure efficiency, here's a graph comparison on the DM side.
image

itsmeow:
Dependency merged. Finishing up caching stuff. It's a bit of a mess, honestly. I wish this module wasn't so opinionated, but that is why it's optional.

dmsrc/iconforge.dm
```diff
@@ -0,0 +1,33 @@
...
+/// list("type" = "BlendIcon", "icon" = sprite_object, "blend_mode" = ICON_OVERLAY)
+/// list("type" = "Scale", "width" = 32, "height" = 32)
+/// list("type" = "Crop", "x1" = 0, "y1" = 0, "x2" = 32, "y2" = 32)
+/// Returns a SpritesheetResult as JSON, containing fields:
+/// list(
+///     "sizes" = list("32x32", "64x64", ...etc)
+///     "sprites" = list("sprite_name" = list("size_id" = "32x32", "position" = 0), ...)
+///     "error" = "[A string, empty if there were no errors.]"
+/// )
+/// In the event of an unrecoverable error, where the spritesheet could not even generate, returns a string containing the error.
```
ZeWaka, line 25: This line is vague, please be more clear about how it relates to the above code. On first read I thought this meant
`return "couldn't read shit"` essentially

Cargo.toml
```diff
@@ -62,7 +62,9 @@ rayon = { version = "1.8", optional = true }
 dbpnoise = { version = "0.1.2", optional = true }
 pathfinding = { version = "4.4", optional = true }
 num-integer = { version = "0.1.45", optional = true }
-dmi = { version = "0.3.1", optional = true }
+dmi = { version = "0.3.3", optional = true }
+tracy_full = { version = "1.6.1", optional = true }
+#tracy_full = { version = "1.6.1", optional = true, features = ["enable"] }
```
ZeWaka, line 67: should remove this whenever you're done

src/byond.rs
```diff
@@ -84,3 +90,29 @@ byond_fn!(
         Some(env!("CARGO_PKG_VERSION"))
     }
 );
+
+// Print any panics before exiting.
```
ZeWaka, line 94:
```suggestion
/// Print any panics before exiting.
```

src/iconforge.rs
```diff
@@ -0,0 +1,928 @@
...
+                return Err(format!(
+                    "Invalid dir {} or size of dirs {} in {} state: {} for sprite {}",
+                    icon.dir, state.dirs, icon.icon_file, icon.icon_state, sprite_name
+                ));
+            }
+        };
+        if (state.dirs == 1 && dir != Dirs::SOUTH)
+            || (state.dirs == 4 && !dmi::dirs::CARDINAL_DIRS.contains(&dir))
+            || (state.dirs == 8 && !dmi::dirs::ALL_DIRS.contains(&dir))
+        {
```
ZeWaka, line 640: :sus: didn't i just merge this into dmi

src/iconforge.rs
```diff
@@ -0,0 +1,928 @@
...
+    blend(
+        Rgba::from_array(color),
+        Rgba::from_array(color2),
+        blend_mode,
+    )
+    .into_array()
+}
+
+/// Blends two colors according to blend_mode. The numbers correspond to BYOND blend modes.
+fn blend(color: Rgba, color2: Rgba, blend_mode: u8) -> Rgba {
```
ZeWaka, line 899: not sure on the design of this, to me this seems like it should be on the rgba struct

src/iconforge.rs
```diff
@@ -0,0 +1,928 @@
...
+        Rgba {
+            r: rgb_fn(color.r, color2.r, color.a, color2.a),
+            g: rgb_fn(color.g, color2.g, color.a, color2.a),
+            b: rgb_fn(color.b, color2.b, color.a, color2.a),
+            a: a_fn(color.a, color2.a),
+        }
+    }
+}
+
+fn blend_u8(color: &[u8], color2: &[u8], blend_mode: u8) -> [u8; 4] {
```
ZeWaka, line 889: again, why not on Rgba

src/iconforge.rs
itsmeow: I havent pushed updates yet

itsmeow:
This PR is now feature complete.

Cargo.toml
```diff
@@ -62,7 +62,8 @@ rayon = { version = "1.8", optional = true }
 dbpnoise = { version = "0.1.2", optional = true }
 pathfinding = { version = "4.4", optional = true }
 num-integer = { version = "0.1.45", optional = true }
-dmi = { version = "0.3.1", optional = true }
+dmi = { version = "0.3.4", optional = true }
+tracy_full = { version = "1.6.1", optional = true }
```
ZeWaka, line 66: it's still here

ZeWaka: also failing build lol

itsmeow: There's zone and frame definitions. It's useful to have around, it does noop when features doesn't include enabled since it's just macros.

itsmeow:
This is only failing build because rust added a new lint that affects the hash module.

itsmeow:
OpenDream hast mislead my blending again. 😔

itsmeow:
I have a branch implementing GAGS into iconforge, although I'd prefer it to be part of another PR. Should I just push it here or wait for merge? It's just one commit atm.

dmsrc/iconforge.dm
```diff
@@ -0,0 +1,59 @@
...
+/// list("type" = "Crop", "x1" = 1, "y1" = 1, "x2" = 32, "y2" = 32) // (BYOND icons index from 1,1 to the upper bound, inclusive)
+///
+/// Returns a SpritesheetResult as JSON, containing fields:
+/// list(
+///     "sizes" = list("32x32", "64x64", ...),
+///     "sprites" = list("sprite_name" = list("size_id" = "32x32", "position" = 0), ...),
+///     "dmi_hashes" = list("icons/path_to/an_icon.dmi" = "d6325c5b4304fb03", ...),
+///     "sprites_hash" = "a2015e5ff403fb5c", // This is the xxh64 hash of the INPUT field "sprites".
+///     "error" = "[A string, empty if there were no errors.]"
+/// )
```
Cyprex, line 33: I'm a bit confused here- this says it can return multiple sizes- I was under the impression this would return a single file, which can only have sprites of a single size.

itsmeow: Nope. The DM implementation of spritesheets works the same way. One asset datum can hold multiple sets of css and png files for a size. Only one png for each size, just multiple files.

optimumtact:
@ZeWaka is there a reason this isn't merge yet

ZeWaka:
> @ZeWaka is there a reason this isn't merge yet

Waiting on me to do a final review early this week per my comment yesterday in tooling questions

src/hash.rs
```diff
@@ -71,16 +71,21 @@ fn hash_algorithm<B: AsRef<[u8]>>(name: &str, bytes: B) -> Result<String> {
             hasher.write(bytes.as_ref());
             Ok(format!("{:x}", hasher.finish()))
         }
+        "xxh64_fixed" => {
+            let mut hasher = XxHash64::with_seed(17479268743136991876);
```
ZeWaka, line 75: seed?

src/iconforge.rs
```diff
@@ -0,0 +1,1088 @@
...
+    if !errors.lock().unwrap().is_empty() {
+        return Err(errors.lock().unwrap().join("\n"));
+    }
+    Ok(())
+}
+
+/// Converts an IO icon to one with icon_hash_input
+fn icon_from_io(icon_in: IconObjectIO) -> IconObject {
+    zone!("icon_from_io");
+    // TODO: can probably convert this function to just lazily attaching icostring to a RefCell<> or something
```
ZeWaka, line 648: todo?

src/iconforge.rs
```diff
@@ -0,0 +1,1088 @@
...
+                    blend_mode,
+                },
+                TransformIO::Crop { x1, y1, x2, y2 } => Transform::Crop { x1, y1, x2, y2 },
+                TransformIO::Scale { width, height } => Transform::Scale { width, height },
+            })
+            .collect(),
+        transform_hash_input: String::new(),
+        icon_hash_input: String::new(),
+    };
+    result.gen_icon_hash_input().unwrap(); // unsafe but idc
```
ZeWaka, line 674: ?

dmsrc/iconforge.dm
```diff
@@ -0,0 +1,59 @@
...
+/// In the case of an unrecoverable panic from within Rust, this function ONLY returns a string containing the error.
+#define rustg_iconforge_generate(file_path, spritesheet_name, sprites, hash_icons) RUSTG_CALL(RUST_G, "iconforge_generate")(file_path, spritesheet_name, sprites, "[hash_icons]")
+/// Returns a job_id for use with rustg_iconforge_check()
+#define rustg_iconforge_generate_async(file_path, spritesheet_name, sprites, hash_icons) RUSTG_CALL(RUST_G, "iconforge_generate_async")(file_path, spritesheet_name, sprites, "[hash_icons]")
+/// Returns the status of a job_id
+#define rustg_iconforge_check(job_id) RUSTG_CALL(RUST_G, "iconforge_check")("[job_id]")
```
ZeWaka, line 39: Are you just unable to get a `SpritesheetResult`? Unclear.

Cargo.toml
ZeWaka, line 66:
```suggestion
tracy_full = { version = "1.7.1", optional = true }
```

ZeWaka:
i would update the build fails but you are cringe and don't let me push to your branch

itsmeow:
sorries

src/hash.rs
itsmeow: Yes. It's a fixed hash, so the seed needs to be the same for all hashes. This is used for caching checks in iconforge

src/iconforge.rs
itsmeow: Yeah I'm not really sure how to do this and it works, it's just a potential microop for memory and processing to clone stuff

src/iconforge.rs
itsmeow: The unwrap() could technically panic. I can add a fail check.

dmsrc/iconforge.dm
itsmeow: Nope. It's the same as all other rustg async implementations. The comment is actually more than most get. It returns the job status, which is a list of DEFINEs that all projects already have.

src/hash.rs
ZeWaka: why this number

dmsrc/iconforge.dm
ZeWaka: That doesn't make sense to me. `http_request_async` does not work this way, the `_check` will return the result if it is complete. That's also what `jobs::check` does internally, which is what is used here.

itsmeow: Then I am remembering wrong. It does that, then

src/hash.rs
itsmeow: I picked it. It doesn't have any meaning.

src/iconforge.rs
```diff
@@ -0,0 +1,1088 @@
...
+                apply_all_transforms(&mut other_image, &icon.transform)?;
+            };
+            for x in 0..std::cmp::min(image.width(), other_image.width()) {
+                for y in 0..std::cmp::min(image.width(), other_image.width()) {
+                    let px1 = image.get_pixel_mut(x, y);
+                    let px2 = other_image.get_pixel(x, y);
+                    let pixel_1 = px1.channels();
+                    let pixel_2 = px2.channels();
+
+                    let blended = Rgba::blend_u8(pixel_1, pixel_2, *blend_mode);
```
Cyprex, line 882: I don't *think* this way of looping is particularly amenable to performance- blend_mode is a loop invariant so this seems suitable for SIMD, but because the switch for blend_mode happens later down the line for each pixel the compiler may have a hard time taking advantage of that. Imo the switch for blend_mode should be hoisted outside of this loop, so that the actual pixel operations can be a tight loop. That is assuming the preceding body of this loop obtaining the pixel values is very cheap, that the actual blending perf even matters!

dmsrc/iconforge.dm
ZeWaka: Then you should update the documentation - that's my main concern.

src/hash.rs
ZeWaka:
```suggestion
            let mut hasher = XxHash64::with_seed(17479268743136991876); // random number
```

src/iconforge.rs
itsmeow: You are correct, good catch! I can peek at the assembly and see if the compiler actually inlines the blend function. It would probably make the code much messier though, maybe I can use a macro or something.

itsmeow: If you look at the tracy profiling, most of the time is spent parsing DMI files from the disk, but icon blending is still a pretty decent portion, especially for spritesheets like preferences which use a lot of operations. The main limitation is actually IO which is why everything is batched into groups and multithreaded where possible. Pretty light on actual computation.

ZeWaka:
https://github.com/BeeStation/BeeStation-Hornet/pull/10404
https://github.com/BeeStation/BeeStation-Hornet/pull/10615

src/iconforge.rs
itsmeow: I don't really have a lot of time to deeply investigate the performance implications of this at the moment

AffectedArc07:
Cannot wait for an implementation of this, please @ me with the PR when its done.

itsmeow:
> Cannot wait for an implementation of this, please @ me with the PR when its done.

On TG?
I've already implemented it on Bee, the PRs are mentioned above.

itsmeow:
One more relevant PR if anyone was going to port btw:
https://github.com/BeeStation/BeeStation-Hornet/pull/10649

PR #163, opened 2023-12-23
pathfinder: stringify args expecting numbers

vvvv-vvvv:
I cba to test but I believe 515 changed some behavior related to this


ZeWaka:
tgmc is the one who uses this anyways, so you probably know best

PR #167, opened 2024-04-11
Routine Update PR

ZeWaka:
title, what else do you expect

some version updates, version our own mysql
mysql maintainer only releases a new version once a year, they (really ring) broke compilation on i686, etc. - easier to just vendor our own


MrStonedOne (approved):
based

PR #168, opened 2024-04-15
add random_replacetext, a replacement for stars()

Kapu1178:
Adds a new function, `random_replacetext()`, which randomly replaces non-whitespace characters in a string with another character. This is to replace the `stars()` function found in every SS13 codebase, which is awfully, awfully slow due to byond's string interning.

Please ignore the overtime values for this, as I simply ran each proc 10,000 times in the same tick.
image

### DIFF
--- a/dmsrc/random_replacetext.dm
+++ b/dmsrc/random_replacetext.dm
@@ -1,0 +1,1 @@
+#define rustg_random_replacetext(text, probability, replacement) RUSTG_CALL(RUST_G, "random_replacetext")(text, probability, replacement)
--- a/src/random_replacetext.rs
+++ b/src/random_replacetext.rs
@@ -1,5 +1,5 @@
-use rand::Rng;
 use crate::error::Result;
+use rand::Rng;
 
 byond_fn!(fn random_replacetext(text, prob, rand_char) {
     replacetext(text, prob, rand_char).ok()
@@ -18,8 +18,7 @@ fn replacetext(text: &str, prob_as_str: &str, replacement_str: &str) -> Result<S
 
         if rng.gen_ratio(prob, 100) {
             string_return.push_str(replacement_str);
-        }
-        else {
+        } else {
             string_return.push(character);
         }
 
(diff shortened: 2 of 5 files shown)


dmsrc/random_replacetext.dm
```diff
@@ -0,0 +1 @@
+#define rustg_random_replacetext(text, probability, replacement) RUSTG_CALL(RUST_G, "random_replacetext")(text, probability, replacement)
```
Kapu1178, line 1:
```suggestion
#define rustg_random_replacetext(text, probability, replacement) RUSTG_CALL(RUST_G, "random_replacetext")(text, "[probability]", replacement)
```

ZeWaka (commented):
i don't think this is worth the compile time of 2 seconds max, much less the maintainer overhead and me having to write this sentence

Kapu1178:
> i don't think this is worth the compile time of 2 seconds max, much less the maintainer overhead and me having to write this sentence

So, let's put into perspective why this is useful. This proc, on TG, is labelled as so expensive you should never use it. It used to be a crash vector on Paradise. This proc used to be used alot, but was scrubbed away because of the cost it incurred. Besides, this code is braindead simple (because im braindead and I wrote it), maintaining it isn't exactly a challenge.

src/random_replacetext.rs
```diff
@@ -0,0 +1,26 @@
+use crate::error::Result;
+use rand::Rng;
+
+byond_fn!(fn random_replacetext(text, prob, rand_char) {
+    replacetext(text, prob, rand_char).ok()
+});
+
+fn replacetext(text: &str, prob_as_str: &str, replacement_str: &str) -> Result<String> {
+    let prob = prob_as_str.parse::<u32>()?;
```
ZeWaka, line 9: Have you considered how this error will propagate up?

ZeWaka (changes requested):
blocking: feature not documented properly in code nor in readme

if you can get another maintainer not from daedalus to care about this we can merge it

PR #171, opened 2024-04-21
Adds `sanitize_html`, a whitelist based HTML sanitizer.

Kapu1178:
Adds a customizable HTML sanitizer function using the Ammonia crate. Out of the box, it will:
* Strip <script> and <style> attributes, as well as their contents.
* Prune all URL schemes, including byond://
* Prune all HTML attributes and CSS tags, but not their contents.

By providing json encoded lists, you can whitelist given attributes or tags to not be pruned. I have included a curated tag list in the dm source file for this module that will whitelist most safe CSS attributes.

It occured to me that alot of servers run things like old papercode, which does not sanitize on the server side before being viewable by a client. Sanitizing strings with DM would be an absolute performance nuke, assuming you could even make it bulletproof in the first place.
Here is a recommended default tag whitelist
```
list(
	"b","br",
	"center", "code",
	"dd", "del", "div", "dl", "dt",
	"em",
	"font",
	"h1", "h2", "h3", "h4", "h5", "h6", "hr",
	"i", "ins",
	"li",
	"menu",
	"ol",
	"p", "pre",
	"span", "strong",
	"table",
	"tbody",
	"td",
	"th",
	"thead",
	"tfoot",
	"tr",
	"u",
	"ul",
)

```


Kapu1178:
>  Error: "sanitize = [\"ammonia\", \"maplit\", \"serde_json\"] is not sorted in Cargo.toml default features"
I am unsure how to fix this.

src/sanitize.rs
```diff
@@ -0,0 +1,33 @@
...
+    tag_whitelist_json: &str,
+) -> Result<String> {
+    let attribute_whitelist: HashSet<&str> = serde_json::from_str(attribute_whitelist_json)?;
+    let tag_whitelist: HashSet<&str> = serde_json::from_str(tag_whitelist_json)?;
+
+    let mut prune_url_schemes = ammonia::Builder::default().clone_url_schemes();
+    prune_url_schemes.insert("byond");
+
+    let sanitized = ammonia::Builder::empty()
+        .clean_content_tags(hashset!["script", "style"]) // Completely forbid script and style attributes.
```
Mothblocks, line 24: You don't need maplit for this.

```rs
HashSet::from_iter(["script", "style"])
```

dmsrc/sanitize.dm
```diff
@@ -0,0 +1,36 @@
+/*
+ * Takes in a string and json_encode()"d lists to produce a sanitized string.
+ * This function operates on whitelists, there is currently no way to blacklist.
+ * Args:
+ * * text: the string to sanitize.
+ * * attribute_whitelist_json: a json_encode()'d list of HTML attributes to allow in the final string.
+ * * tag_whitelist_json: a json_encode()'d list of HTML tags to allow in the final string.
+ */
+#define rustg_sanitize_html(text, attribute_whitelist_json, tag_whitelist_json) RUSTG_CALL(RUST_G, "sanitize_html")(text, attribute_whitelist_json, tag_whitelist_json)
```
Mothblocks, line 9: Semicolon?

dmsrc/sanitize.dm
```diff
@@ -0,0 +1,36 @@
+/*
+ * Takes in a string and json_encode()"d lists to produce a sanitized string.
+ * This function operates on whitelists, there is currently no way to blacklist.
+ * Args:
+ * * text: the string to sanitize.
+ * * attribute_whitelist_json: a json_encode()'d list of HTML attributes to allow in the final string.
+ * * tag_whitelist_json: a json_encode()'d list of HTML tags to allow in the final string.
```
Mothblocks, line 7: The interface should take a list and json_encode in itself.

Kapu1178: I didn't do this so that you can store pre-encoded global lists to save on perf.

dmsrc/sanitize.dm
Kapu1178: I don't understand, am I missing something here?

dmsrc/sanitize.dm
Cyprex: Wouldn't that mean it's encoding on every call? The thing is that this will likely be called many times with only one or a few lists, so this introduces extra overhead.

src/sanitize.rs
```diff
@@ -0,0 +1,32 @@
...
+
+    let mut prune_url_schemes = ammonia::Builder::default().clone_url_schemes();
+    prune_url_schemes.insert("byond");
+
+    let sanitized = ammonia::Builder::empty()
+        .clean_content_tags(HashSet::from_iter(["script", "style"])) // Completely forbid script and style attributes.
+        .link_rel(Some("noopener")) // https://mathiasbynens.github.io/rel-noopener/
+        .url_schemes(prune_url_schemes)
+        .generic_attributes(attribute_whitelist)
+        .tags(tag_whitelist)
```
vuonojenmustaturska, line 27: wouldn't it make sense to keep this around rather than build it anew on every invocation

Kapu1178: I'd have to hash the arguments and such and that's out of my skill set presently.

dmsrc/sanitize.dm
```diff
@@ -0,0 +1,36 @@
...
+ */
+#define rustg_sanitize_html(text, attribute_whitelist_json, tag_whitelist_json) RUSTG_CALL(RUST_G, "sanitize_html")(text, attribute_whitelist_json, tag_whitelist_json)
+
+/*
+* Here is a recommended default tag whitelist
+list(
```
ZeWaka, line 13: This isn't going to turn out super well due to this all being shoved into a single DM file

optimumtact:
looks about right :+2:

Kapu1178:
mods? mergies? @ZeWaka

PR #174, opened 2024-05-27
Adds batchnoise to the default features set

GoldenAlpharex:
Simply adds `batchnoise` to the default feature set, as I'm going to need to use it with biome generation on /tg/.

I don't know if I need to do anything else, I don't think that I do, but do let me know in case there IS something else I forgot to do.

PR'd with approval from @LemonInTheDark and @ZeWaka.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,6 +68,7 @@ tracy_full = { version = "1.7.1", optional = true }
 [features]
 default = [
     "acreplace",
+    "batchnoise",
     "cellularnoise",
     "dmi",
     "file",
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -81,6 +81,7 @@ default = [
     "time",
     "toml",
     "url",
+    "batchnoise",
```
Mothblocks, line 84: abc order

GoldenAlpharex: Good catch, my bad. Was following the below list, but it makes a lot of sense now that I look at it.

ZeWaka:
the arrangement in the All list needs to be fixed but ill do that in another PR

PR #178, opened 2024-07-03
Fast poisson sampling

FluffyGhoster:
Hello!

I have backported to upstream (you) our (Aurora) addition of a fast poisson sampling function used to get noise samples for random generation

I'm not familiar with rust, but I have tested it and it compiles successfully, we also use this on live since a long while and it seems to work well

Thank you for the consideration!


dmsrc/noise.dm
```diff
@@ -1 +1,2 @@
 #define rustg_noise_get_at_coordinates(seed, x, y) RUSTG_CALL(RUST_G, "noise_get_at_coordinates")(seed, x, y)
+#define rustg_noise_poisson_sample(seed, x, y, r) RUSTG_CALL(RUST_G, "generate_poisson_sample")(seed, x, y, r)
```
ZeWaka, line 2: Would be nice to have some documentation here

ZeWaka: Especially since it seems to be batched

FluffyGhoster: Batched?

ZeWaka: it looks like it returns a list of data, not a single point

FluffyGhoster: So I have asked to the original author of this function for an outline of how it works, and this is his reply (paraphrased as documentation):

`it works the same way as batched noise generation in that it returns a string of 1s and 0s - you give it a seed, the X and Y size of the map you want, and the radius of the samples (i.e., how far away each one should be from the next one), and you get back a string that is X*Y long; it works the same way dbp_generate does but with fewer params because it's poisson instead of batched perlin`

In documentation form:

```
params:
  seed: str
  x: int, width of the noisemap
  y: int, height of the noisemap
  radius: int, distance between points on the noisemap

returns:
  string: a X*Y length string of 1s and 0s representing a 2D poisson sample collapsed into a 1D string
```

ZeWaka: Yeah, x and y are misleading because those would generally imply the coordinates. Probably should change to w and l - but yeah you should add the documentation - unless you don't know how to, then I can help :)

FluffyGhoster: I don't know how to add the documentation, how would I do it?

ZeWaka: i got chu fam

ZeWaka (approved):

PR #180, opened 2024-07-12
Add method of parsing revdate for HEAD directly from logs

ZephyrTFA:


src/git.rs
```diff
@@ -24,3 +29,19 @@ byond_fn!(fn rg_git_commit_date(rev, format) {
         Some(datetime.format(format).to_string())
     })
 });
+
+byond_fn!(fn rg_git_commit_date_head(format) {
+    let head_log_path = Path::join(&PathBuf::from_str(".git").ok()?, "logs").join("HEAD");
```
ZeWaka, line 34: why not
```rs
let head_log_path = Path::new(".git")
        .join("logs")
        .join("HEAD");
```

ZephyrTFA: > why not
> ```rs
> let head_log_path = Path::new(".git")
>         .join("logs")
>         .join("HEAD");
> ```

great question
```suggestion
let head_log_path = Path::new(".git")
	.join("logs")
	.join("HEAD");
```

PR #182, opened 2024-07-15
Compiling error reworks. allow 64 with feature, block newer rustc without a feature.

MrStonedOne:
Adds features that to allow compiling under 64 bit or a rustc version that is too new for windows 7 and 8 (note, the game servers are still windows server 2012, so rustc dropping windows 8 support under 1.76 will also fuck prod.)

Compiling under 64bit is required for OD support and this has turned from an error that was suppose to prevent footguns into an error that is itself a footgun.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -12,7 +12,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: dtolnay/rust-toolchain@1.75
         with:
           targets: i686-pc-windows-msvc
           components: rustfmt, clippy
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2958,6 +2958,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "rustversion"
+version = "1.0.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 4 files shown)


ZeWaka:
While this is good/necessary, we should probably bold / emphasize the bit in the readme even more so less people fall for the issue.

I don't see an easy way to emit compiler info statements unfortunately. See, this issue which is maybe familiar: https://internals.rust-lang.org/t/pre-rfc-add-compile-warning-macro/9370/8 and  https://github.com/rust-lang/rust/issues/54140

Cyberboss:
Is it possible to feature flag gate the error?

ZeWaka:
> Is it possible to feature flag gate the error?

```toml
[features]
allow_non_32bit = []
```
```rs
#[cfg(all(not(target_pointer_width = "32"), not(feature = "allow_non_32bit")))]
compile_error!("Compiling for non-32bit is not allowed without enabling the `allow_non_32bit` feature.");
```

Cyberboss (changes requested):
What waka said

MrStonedOne:
Added features that block compiling under 64 bit or a rustc version that is too new for windows 7 and 8 (note, the game servers are still windows server 2012, so rustc dropping windows 8 support under 1.76 will also fuck us.

MrStonedOne:
This will be easier to copy and paste the windows 7 bits into a new webedit then to deal with conflicts, half of which are for lines that will no longer be changed.

PR #183, opened 2024-07-20
Windows 7

Mothblocks:
Upgrades all Cargo packages, plus replaces reqwest with ureq because reqwest relies on tokio, which relies on an old version of mio, which relies on a specific version of windows-targets, which whatever it is it didn't work

Doesn't do anything about the 32/64 bit thing, someone else can do that in a separate PR

Solves windows 7 part of #182

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -17,7 +17,7 @@ jobs:
           targets: i686-pc-windows-msvc # Used for clippy, since it doesn't need to build the entire std
           components: rustfmt, clippy, rust-src
 
-      # - uses: Swatinem/rust-cache@v2
+      - uses: Swatinem/rust-cache@v2
 
       - name: Clippy (all features)
         run: cargo clippy --target i686-pc-windows-msvc --features all --locked -- -D warnings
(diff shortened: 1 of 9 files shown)


.github/workflows/rust.yml
```diff
@@ -26,14 +26,14 @@ jobs:
         run: cargo fmt -- --check
 
       - name: Build (release) (default features)
-        run: cargo build --target i686-pc-windows-msvc --locked --release
+        run: cargo build -Z build-std --target i686-win7-windows-msvc --locked --release
```
ZeWaka, line 29: For those unaware of why we are building std and thus using a nightly toolchain,
https://github.com/rust-lang/rust/pull/118150

ZeWaka, line 29: We should probably add instructions to the readme regarding building for windows 7.

a la
https://github.com/rust-lang/rust/blob/master/src/doc/rustc/src/platform-support/win7-windows-msvc.md

ZeWaka (changes requested):
req changes re: readme

ZeWaka (approved):

PR #184, opened 2024-08-09
Allow compiling non-32bit under feature flag

MrStonedOne:
Adds feature flag to allow compiling under 64 bit for OD reasons.


ZeWaka (approved):

MrStonedOne:
@Mothblocks could you check out this travis error?

PR #188, opened 2024-09-16
IconForge: GAGS

itsmeow:
Re-implements GAGS config loading and icon generation using IconForge, can basically 1:1 replace any GAGS generate() tasks in DM and can be up to 250x faster per unique config-color combination. Worst case goes from 250ms to 1ms.

Works by outputting the result to a DMI in `tmp` and assigning it to `icon` in DM.

This version of GAGS **does not support** (it will soft fail and just not generate the icon correctly):
- The `color_matrix` layer type
- The `or` blending mode
- BYOND Parity on obscure operations with multiple icon_states at the same time (animated sprites, unscoped/stateless icons, mixed-dirs aka overlaying icons with different dir amounts)

image

A DM implementation can be found in this PR: https://github.com/BeeStation/BeeStation-Hornet/pull/10455

This has been in active use on prod since January 2024 with zero issues.

This works to address https://github.com/tgstation/dev-cycles-initiative/issues/9 if it were to be implemented on TG


optimumtact:
what is blocking this

ZeWaka (approved):

PR #192, opened 2024-11-09
Feature: rustg_sound_length()

Kapu1178:
Adds a function for getting the duration of an audio file.
Also adds a function for getting the durations of a list of audio files.

TODO:
- [x] Add a version that accepts a list of filepaths instead.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,7 +64,9 @@ ammonia = { version = "4.0.0", optional = true }
 fast_poisson = { version = "0.5.2", optional = true, features = [
     "single_precision",
 ] } # Higher versions have problems with x86 due to 'kiddo'.
-symphonia = { version = "0.5.4", optional = true }
+symphonia = { version = "0.5.4", optional = true, features = [
+    "all-codecs",
+] }
 
 [features]
 default = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,6 +64,7 @@ ammonia = { version = "4.0.0", optional = true }
 fast_poisson = { version = "0.5.2", optional = true, features = [
     "single_precision",
 ] } # Higher versions have problems with x86 due to 'kiddo'.
+symphonia = { version = "0.5.4", optional = true }
 
 [features]
 default = [
@@ -79,6 +80,7 @@ default = [
     "noise",
     "rustls_tls",
     "sanitize",
+    "sound_len",
     "sql",
     "time",
     "toml",
@@ -98,6 +100,7 @@ all = [
     "noise",
     "rustls_tls",
     "sanitize",
+    "sound_len",
     "sql",
     "time",
     "toml",
@@ -123,6 +126,7 @@ http = ["ureq", "serde", "serde_json", "once_cell", "jobs"]
 json = ["serde", "serde_json"]
 log = ["chrono"]
 sanitize = ["ammonia", "serde_json"]
+sound_len = ["symphonia"]
 sql = ["mysql", "serde", "serde_json", "once_cell", "dashmap", "jobs"]
 time = []
 toml = ["serde", "serde_json", "toml-dep"]
--- a/dmsrc/sound_len.dm
+++ b/dmsrc/sound_len.dm
@@ -15,8 +15,9 @@
 		else
 			CRASH("rustg_sound_length does not support non-static file refs.")
 
-	if(!isnull((. = sound_cache[file_path])))
-		return .
+	var/cached_length = sound_cache[file_path]
+	if(!isnull(cached_length))
+		return cached_length
 
 	var/ret = RUSTG_CALL(RUST_G, "sound_len")(file_path)
 	var/as_num = text2num(ret)
--- a/dmsrc/sound_len.dm
+++ b/dmsrc/sound_len.dm
@@ -1,3 +1,4 @@
+// Provided a static RSC file path or a raw text file path, returns the duration of the file in deciseconds as a float.
 /proc/rustg_sound_length(file_path)
 	var/static/list/sound_cache
 	if(isnull(sound_cache))
@@ -25,3 +26,8 @@
 
 	sound_cache[file_path] = as_num
 	return as_num
+
+#define RUSTG_SOUNDLEN_SUCCESSES "successes"
+#define RUSTG_SOUNDLEN_ERRORS "errors"
+// Returns a list of lists "successes" and "errors". Successes are file_path : duration. Errors are file_path : error.
+#define rustg_sound_length_list(file_paths) json_decode(RUSTG_CALL(RUST_G, "sound_len_list")(json_encode(file_paths)))
--- a/dmsrc/sound_len.dm
+++ b/dmsrc/sound_len.dm
@@ -1,4 +1,4 @@
-// Provided a static RSC file path or a raw text file path, returns the duration of the file in deciseconds as a float.
+/// Provided a static RSC file path or a raw text file path, returns the duration of the file in deciseconds as a float.
 /proc/rustg_sound_length(file_path)
 	var/static/list/sound_cache
 	if(isnull(sound_cache))
@@ -10,7 +10,7 @@
 		if(!isfile(file_path))
 			CRASH("rustg_sound_length error: Passed non-text object")
 
-		if(length("[file_path]"))
+		if(length("[file_path]")) // Runtime generated RSC references stringify into 0-length strings.
 			file_path = "[file_path]"
 		else
 			CRASH("rustg_sound_length does not support non-static file refs.")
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -123,6 +126,7 @@ http = ["ureq", "serde", "serde_json", "once_cell", "jobs"]
 json = ["serde", "serde_json"]
 log = ["chrono"]
 sanitize = ["ammonia", "serde_json"]
+sound_len = ["symphonia"]
```
ZeWaka, line 129: add to readme

Cargo.toml
```diff
@@ -79,6 +80,7 @@ default = [
     "noise",
     "rustls_tls",
     "sanitize",
+    "sound_len",
```
ZeWaka, line 83: no

dmsrc/soundlen.dm
```diff
@@ -0,0 +1,27 @@
+/proc/rustg_sound_length(file_path)
```
ZeWaka, line 1: documentation

src/sound_len.rs
```diff
@@ -0,0 +1,94 @@
...
+    // Try to decode the data packet
+    let decoded_packet = match decoder.decode(&encoded_packet) {
+        Ok(r) => r,
+        Err(_e) => return Ok(String::from("Failed to decode packet!"))
+    };
+
+    // Grab the sample rate from the spec of the buffer.
+    let sample_rate: f64 = decoded_packet.spec().rate as f64;
+    // Math!
+    let duration_in_seconds: f32 = (samples_capacity / sample_rate) as f32;
```
ZeWaka, line 91: all of this seems way overkill there has to be a way to do it in the library easily

Cargo.toml
Kapu1178: tgstation sound maintainer wants this feature so it makes sense to me

src/sound_len.rs
Kapu1178: added a simpler version and kept the codec reader as a fallback

dmsrc/sound_len.dm
```diff
@@ -0,0 +1,33 @@
...
+		if(!isfile(file_path))
+			CRASH("rustg_sound_length error: Passed non-text object")
+
+		if(length("[file_path]"))
+			file_path = "[file_path]"
+		else
+			CRASH("rustg_sound_length does not support non-static file refs.")
+
+	if(!isnull((. = sound_cache[file_path])))
+		return .
```
Mothblocks, line 19: This code is doing too much and is not easy to understand, please write it out more directly

```suggestion
	var/cached_length = sound_cache[file_path]
	if(!isnull(cached_length))
		return cached_length
```

src/sound_len.rs
```diff
@@ -0,0 +1,157 @@
...
+    // Gracefully exit if the filepath is invalid.
+    if !path.exists() {
+        return Err(SoundLen(format!("Bad path: {}", sound_path)));
+    }
+
+    // Try to open the file
+    let sound_src = match File::open(path) {
+        Ok(r) => r,
+        Err(_e) => return Err(SoundLen("Couldn't open file!".to_string())),
+    };
```
Mothblocks, line 34: Nit: It's good practice in general to just open it, and check if the error is file not found. 

https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.NotFound

src/sound_len.rs
```diff
@@ -0,0 +1,157 @@
...
+
+    // Gracefully exit if the filepath is invalid.
+    if !path.exists() {
+        return Err(SoundLen(format!("Bad path: {}", sound_path)));
+    }
+
+    // Try to open the file
+    let sound_src = match File::open(path) {
+        Ok(r) => r,
+        Err(_e) => return Err(SoundLen("Couldn't open file!".to_string())),
```
Mothblocks, line 33: Why don't we give the error with it? Do we think it'll leak some info?

src/sound_len.rs
```diff
@@ -0,0 +1,157 @@
...
+fn sound_length_decode(probed: ProbeResult) -> Result<f64> {
+    let mut format = probed.format;
+
+    let track = match format.default_track() {
+        Some(r) => r,
+        None => return Err(SoundLen("Could not get default track".to_string())),
+    };
+
+    // Grab the number of frames of the track
+    let samples_capacity: f64 = if let Some(n_frames) = track.codec_params.n_frames {
```
Mothblocks, line 98: Shouldn't need `: f64`?

src/sound_len.rs
Kapu1178: oops, leftover

ZephyrTFA:
holy shit, this will be useful for spatial audio

dmsrc/sound_len.dm
```diff
@@ -0,0 +1,34 @@
+// Provided a static RSC file path or a raw text file path, returns the duration of the file in deciseconds as a float.
+/proc/rustg_sound_length(file_path)
```
ZephyrTFA, line 2: can you add support for frequency overrides?

Kapu1178: No.

dmsrc/sound_len.dm
```diff
@@ -0,0 +1,34 @@
...
+		sound_cache = list()
+
+	. = 0
+
+	if(!istext(file_path))
+		if(!isfile(file_path))
+			CRASH("rustg_sound_length error: Passed non-text object")
+
+		if(length("[file_path]"))
+			file_path = "[file_path]"
```
ZeWaka, line 14: what is this code doing

dmsrc/sound_len.dm
```diff
@@ -0,0 +1,34 @@
+// Provided a static RSC file path or a raw text file path, returns the duration of the file in deciseconds as a float.
```
ZeWaka, line 1: code doc
```suggestion
/// Provided a static RSC file path or a raw text file path, returns the duration of the file in deciseconds as a float.
```

dmsrc/sound_len.dm
```diff
@@ -0,0 +1,34 @@
...
+	if(isnull(ret))
+		. = 0
+		CRASH("rustg_sound_length error: [ret]")
+
+	sound_cache[file_path] = as_num
+	return as_num
+
+#define RUSTG_SOUNDLEN_SUCCESSES "successes"
+#define RUSTG_SOUNDLEN_ERRORS "errors"
+// Returns a list of lists "successes" and "errors". Successes are file_path : duration. Errors are file_path : error.
```
ZeWaka, line 33: code doc and this description is confusing, i don't get the exact data structure

src/sound_len.rs
```diff
@@ -0,0 +1,150 @@
...
+fn get_sound_length_list(list: &str) -> String {
+    let json: Vec<&str> = match serde_json::from_str(list) {
+        Ok(r) => r,
+        Err(_e) => return String::from("Fatal error: Bad json"),
+    };
+
+    let mut successes: HashMap<String, String> = HashMap::new();
+    let mut errors: HashMap<String, String> = HashMap::new();
+
+    for path_string in json.iter() {
```
ZeWaka, line 138: probably should be using parallel iter here, giving up free performance

src/sound_len.rs
```diff
@@ -0,0 +1,150 @@
...
+    Ok(duration_in_desciseconds)
+}
+
+byond_fn!(
+    fn sound_len_list(list) {
+        Some(get_sound_length_list(list))
+    }
+);
+
+fn get_sound_length_list(list: &str) -> String {
```
ZeWaka, line 129: do you really need all of the explicit types inside this function? pretty sure they can be figured out by the compiler fine

src/sound_len.rs
```diff
@@ -0,0 +1,150 @@
...
+    // Try to decode the data packet
+    let decoded_packet = match decoder.decode(&encoded_packet) {
+        Ok(r) => r,
+        Err(e) => return Err(SoundLen(format!("Decode error: {}", e))),
+    };
+
+    // Grab the sample rate from the spec of the buffer.
+    let sample_rate = decoded_packet.spec().rate as f64;
+    // Math!
+    let duration_in_desciseconds = samples_capacity / sample_rate * 10.0;
```
ZeWaka, line 119: parens for math grouping?

ZeWaka (changes requested):

dmsrc/sound_len.dm
Kapu1178: This checks if an RSC path is a real file. If it cant be stringified it isn't a file on the disk.

src/sound_len.rs
Kapu1178: I don't know what this means

src/sound_len.rs
Kapu1178: Clippy told me the parens were unnecesary

dmsrc/sound_len.dm
ZeWaka: then comment that

src/sound_len.rs
ZeWaka: eh whatever you can always change it if you want it to be faster

ZeWaka (approved):

PR #196, opened 2024-12-28
Add hash and iconforge as default features

itsmeow:
Adds `hash` and `iconforge` as default features so they can be used on /tg/ including in the CI process (since embedding an 82MB .so file into the repo is unlikely, and building the binary in CI is slow)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,7 +76,9 @@ default = [
     "dmi",
     "file",
     "git",
+    "hash",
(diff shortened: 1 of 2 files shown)


ZeWaka:
regarding `iconforge`:
image

requested `hash` in maintainerbus

itsmeow:
Approval acquired

image
image - the statement about OpenSSL is incorrect

image

PR #197, opened 2024-12-29
Make iconforge call create_dir_all, flush, and sync_all to deal with file system caches

tigercat2000:
In some rare cases, filesystem caches can cause the DM side of iconforge to experience the rust library returning, before the file is actually available for hashing and fcopy_rsc. This breaks a lot of assumptions that completely derails the code and breaks the asset registration completely.

To avoid this, we use our own bufwriter so we can call flush(), and sync_all().

Also calls create_dir_all as that was also an issue in testing.


ZeWaka:
@itsmeow please advise

itsmeow:
Just need to run a tracy profile on this and compare the performance and it's good with me

ShadowLarkens:
Added another fix for another bug I found: `for y in 0..std::cmp::min(image.width(), other_image.width())` only works for square icons. Using height works for both square and non-square icons.

itsmeow:
So turns out this is way horrible for performance and is likely related to having massive spritesheet file sizes and no page cache on the Linux runners? I'm not sure if merging this is a good idea as they work fine for spritesheets that aren't massive and this adds seconds of overhead to small spritesheets that don't have issues. This needs to check for a certain size (what that size is, I have no idea)

itsmeow (approved):

ZeWaka (commented):
Just checking in, is all this code still needed?

itsmeow:
For some use cases yes. It adds a little bit of overhead, but it's acceptable. The bug fix is necessary.

PR #202, opened 2025-01-19
64 bit lib detection

ZeWaka:
for opendream

### DIFF
--- a/dmsrc/main.dm
+++ b/dmsrc/main.dm
@@ -19,21 +19,25 @@
 /* This comment bypasses grep checks */ /var/__rust_g
 
 /proc/__detect_rust_g()
+	var/__arch_suffix = null
+	#ifdef OPENDREAM
+	__arch_suffix = "64"
+	#endif
 	if (world.system_type == UNIX)
-		if (fexists("./librust_g.so"))
+		if (fexists("./librust_g[__arch_suffix].so"))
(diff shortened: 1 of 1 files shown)


amylizzle:
voicing my support for `librust_g64.so` and `rust_g64.dll`

dmsrc/main.dm
```diff
@@ -19,21 +19,25 @@
 /* This comment bypasses grep checks */ /var/__rust_g
 
 /proc/__detect_rust_g()
+	var/__arch_suffix = null
```
alexkar598, line 22: Underscore prefixing is not necessary within the scope of an underscore prefixed function

alexkar598 (changes requested):
The name is fine

dmsrc/main.dm
ZeWaka: fair

alexkar598 (approved):

Absolucy:
tg already uses `librustg_64.dll` and `rust_g64.dll` anyways: https://github.com/tgstation/tgstation/pull/87313

ZeWaka:
> tg already uses `librustg_64.dll` and `rust_g64.dll` anyways: [tgstation/tgstation#87313](https://github.com/tgstation/tgstation/pull/87313)

Yes, that's why I chose these names, so they could stop being out of compliance with the rustg.dm

PR #210, opened 2025-03-22
Add support for timing out HTTP calls

Cyberboss:
/tg/ TTS is fucked and hanging requests which is blocking shutdown.

Add support for optional timeouts


ZeWaka (commented):
I assume you want this to go out in 3.x and not 4.x, correct?

Cyberboss:
Yeah it _should_ be backwards compatible, assuming serde doesn't error on missing JSON fields

ZeWaka (approved):
this really should be documented in http.dm but nothing is currently and this is somewhat urgent

PR #214, opened 2025-05-02
Adds a new time function, `formatted_timestamp`

Absolucy:
This adds a simple function, `formatted_offset`

It is basically just a wrapper around [chrono's formatting](https://docs.rs/chrono/0.4.41/chrono/format/strftime/index.html), and takes two args: a format, and an optional offset

If the offset is blank, then the system's local time will be used, in order to stay consistent with the behavior of BYOND's `time2text` - otherwise, an offset from UTC in hours can be given.

basically this is better `time2text`, with subsecond support and such

### DIFF
--- a/dmsrc/time.dm
+++ b/dmsrc/time.dm
@@ -1,6 +1,8 @@
 #define rustg_time_microseconds(id) text2num(RUSTG_CALL(RUST_G, "time_microseconds")(id))
 #define rustg_time_milliseconds(id) text2num(RUSTG_CALL(RUST_G, "time_milliseconds")(id))
 #define rustg_time_reset(id) RUSTG_CALL(RUST_G, "time_reset")(id)
+#define rustg_formatted_timestamp(format) RUSTG_CALL(RUST_G, "formatted_timestamp")(format)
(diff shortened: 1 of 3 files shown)


ZeWaka:
I need to fix the stuff in master

dmsrc/time.dm
```diff
@@ -1,6 +1,8 @@
 #define rustg_time_microseconds(id) text2num(RUSTG_CALL(RUST_G, "time_microseconds")(id))
 #define rustg_time_milliseconds(id) text2num(RUSTG_CALL(RUST_G, "time_milliseconds")(id))
 #define rustg_time_reset(id) RUSTG_CALL(RUST_G, "time_reset")(id)
+#define rustg_formatted_timestamp(format) RUSTG_CALL(RUST_G, "formatted_timestamp")(format)
+#define rustg_formatted_timestamp_tz(format, offset) RUSTG_CALL(RUST_G, "formatted_timestamp")(format, offset)
```
ZeWaka, line 5: Needs documentation, I'm guessing this is a UTC offset, but that might not be clear to everyone, especially the formatting

ZeWaka (changes requested):

ZeWaka (approved):

PR #215, opened 2025-05-02
fixes the clippy lints on nightly

Absolucy:
self-explanatory. just so CI works again.


Absolucy:
(i forgot to do `--all-features`, one sec)

PR #216, opened 2025-05-03
Adds `roll_dice`, an advanced xdy dice roller.

ComradeNiobe:
This PR adds a very simple function, `roll_dice`. It is essentially a straight upgrade from BYOND's built-in roll() proc that can do some pretty useful stuff for SS13 servers or other BYOND games that want to go more into RPG-style gameplay, like exploding dice, keeping/dropping highest or lowest, etc.

The actual syntax and examples can be found here: 
https://docs.rs/caith/latest/caith/#syntax 
https://docs.rs/caith/latest/caith/#examples

also this is my first time coding with rust in awhile so any advice is appreciated :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,11 @@
 
 /target
 **/*.rs.bk
+tests/dm/*.dmb
+tests/dm/*.lk
+tests/dm/*.int
+tests/dm/*.rsc
+tests/dm/tmp/
+tests/dm/*.dll
+tests/dm/*.so
+tests/dm/rust_g.dm
--- a/src/url.rs
+++ b/src/url.rs
@@ -24,7 +24,7 @@ fn decode(string: &str) -> Result<String> {
 
 // From `url` crate.
 /// Replace b'+' with b' '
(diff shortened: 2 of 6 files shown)


dmsrc/dice.dm
```diff
@@ -0,0 +1,11 @@
...
+ * Syntax Guide: https://docs.rs/caith/latest/caith/#syntax
+ * Examples: https://docs.rs/caith/latest/caith/#examples
+ *
+ * Args:
+ * * input: the xdy dice to roll; see syntax guide & examples for proper formatting.
+ *
+ * Returns:
+ * * the total sum of the roll as a string.
+ */
+#define rustg_roll_dice(input) RUSTG_CALL(RUST_G, "roll_dice")("[input]")
```
Kapu1178, line 11:
```suggestion
#define rustg_roll_dice(input) text2num(RUSTG_CALL(RUST_G, "roll_dice")("[input]"))
```
text2num returns null so you can still handle error cases.

ZeWaka (changes requested):
Needs to be added to the README

PR #225, opened 2025-07-01
hash: Adds ChaCha20 CSPRNG functions

itsmeow:
Adds to functions to `hash`:

- `rustg_csprng_chacha20`: Cryptographically-secure pseudo-random number generator seeded by the OS/hardware.
- `rustg_prng_chacha20_seeded`: High-quality known-seed deterministic pseudo-random number generator .

Both functions support output into any of the following formats:

- Hex [0-9a-z]
- Alphanumeric [A-Za-z0-9]
- Base64 [A-Za-z0-9+/=]

Both functions take a parameter, `n_bytes` which is the number of bytes sampled by the RNG. The relation of `n_bytes` to string output length is not 1:1 and varies by format.

- **Hex:** `n_bytes * 2`
- **Alphanumeric:** `n_bytes`
- **Base64:** `4 * ceil(n_bytes/3)`


itsmeow:
imo yes, it's using almost all the same packages and the TOTP generator is already there. The only reason I added a CSPRNG is because the TOTP generator needs a cryptographically secure random seed

itsmeow:
Why base32?

Yog is the only current user of this function indexed on GitHub, and [they already convert it to Base32](https://github.com/alexkar598/Yogstation/blob/52a917f1b73605b30408804bcb78df60e5a094a6/code/modules/admin/mfa.dm#L145-L166)

optimumtact:
do the calls block if there is no hardware entropy available?

itsmeow:
> do the calls block if there is no hardware entropy available?

Yes, early in the boot process. The OS will eventually give the necessary data though, from what I understand.
https://docs.rs/getrandom/latest/getrandom/#early-boot

PR #231, opened 2025-07-06
fix http requests not properly setting the status code

Absolucy:
image

the reqwest -> ureq port had a slight whoopsie, non-2xx status code would always error (not setting the `status_code`) field due to how ureq worked, as a non-2xx status code would return an `Err(ureq::Error::Status(status_code, response))` instead of an `Ok(response)`

this fixes that, so that non-2xx status codes will still return the body and such, including setting the status code.

### DIFF
--- a/src/http.rs
+++ b/src/http.rs
@@ -124,11 +124,15 @@ fn construct_request(
 }
 
 fn submit_request(prep: RequestPrep) -> Result<String> {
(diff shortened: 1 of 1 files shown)


ZeWaka:
#205

ZeWaka (commented):
This is also not the way to solve it, it's a flag on the http client or something

optimumtact:
http_real_error_request

PR #234, opened 2025-07-16
DMI metadata injection

Y0SH1M4S73R:
This PR simply adds a function to inject DMI metadata into a PNG file.

In conjunction with `dmi_create_png`'s new support for alpha channels, it will now be possible to create dmis from user-provided pixel data much more performantly than BYOND's own icon operations.

### Example Usage:
```
var/list/pixels = /* Some list containing 4096 rgba strings - 128 rows of 32 pixels, 32 pixels per row, 32 rows per dir. */
rustg_dmi_create_png("output.dmi", "32", "128", pixels)
var/metadata = json_encode(list(\
  "width" = 32,\
  "height" = 32,\
  "states" = list(\
    list(\
      "name" = "",\
      "dirs" = 4,\
    )\
  )\
)
rustg_dmi_inject_metadata("output.dmi", metadata)
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2865,6 +2865,7 @@ dependencies = [
  "regex",
  "serde",
  "serde_json",
+ "serde_repr",
  "sha-1",
  "sha2",
  "symphonia",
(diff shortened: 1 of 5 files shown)


ZeWaka (commented):
seems a bit strange to add a way to inject without an ability to read

Y0SH1M4S73R:
> seems a bit strange to add a way to inject without an ability to read

Well if you want a DMI metadata reading function to go with it, I can do that.

ZeWaka (approved):

PR #236, opened 2025-07-28
IconForge: Headless Icon Generation

itsmeow:
Adds a new headless generation system to IconForge via `rustg_iconforge_generate_headless`:

```dm
/// Creates a single DMI or PNG using 'sprites' as a list of icon states / images.
/// This function is intended for generating icons with only a few states that have little in common with each other, and only one size.
/// For icons with a large number of states, potentially variable sizes, that re-use sets of transforms more than once, or that benefit from caching, use rustg_iconforge_generate.
/// sprites - follows the same format as rustg_iconforge_generate.
/// file_path - the full relative path at which the PNG or DMI will be written. It must be a full filepath such as tmp/my_icon.dmi or my_icon.png
/// flatten - boolean (0 or 1) determines if the DMI output will be flattened to a single frame/dir if unscoped (null/0 dir or frame values).
///
/// Returns a HeadlessResult, decoded to a BYOND list (always, it's not possible for this to panic unless rustg itself has an issue) containing the following fields:
/// list(
///     "file_path" = "tmp/my_icon.dmi" // [whatever you input returned back to you, null if there was a fatal error]
///     "width" = 32 // the width, which is determined by the first entry of 'sprites', null if there was a fatal error
///     "height" = 32 // the height, which is determined by the first entry of 'sprites', null if there was a fatal error
///     "error" = "[A string, null if there were no errors.]"
/// )
#define rustg_iconforge_generate_headless(file_path, sprites, flatten) json_decode(RUSTG_CALL(RUST_G, "iconforge_generate_headless")(file_path, sprites, "[flatten]"))
```

This system is intended for generating icons with only a few states that have little in common with each other, and only one size, as opposed to the traditional generation which is suitable for large batches that re-use icons regularly.

The new function is suitable for creating drop-in replacement for one-time icon operations and could be used for something like generating a minimap with DrawBox. Speed relative to native BYOND icons is not guaranteed due to the I/O cost and less optimization / caching, but it's likely to be faster for icons that are built with a large number of transforms in sequence, because it doesn't resolve the icon until it's converted.

Also fixes a bug where flattened icons could pollute the cache with incorrect data by splitting the caches for flattened and non-flattened icons.

Adds error-tests for icon generation using the new headless generator, checking for common error cases. Also tests headless generation for consistency with BYOND and regular batch generation.

### DIFF
--- a/dmsrc/iconforge.dm
+++ b/dmsrc/iconforge.dm
@@ -46,6 +46,21 @@
 #define rustg_iconforge_generate(file_path, spritesheet_name, sprites, hash_icons, generate_dmi, flatten) RUSTG_CALL(RUST_G, "iconforge_generate")(file_path, spritesheet_name, sprites, "[hash_icons]", "[generate_dmi]", "[flatten]")
 /// Returns a job_id for use with rustg_iconforge_check()
 #define rustg_iconforge_generate_async(file_path, spritesheet_name, sprites, hash_icons, generate_dmi, flatten) RUSTG_CALL(RUST_G, "iconforge_generate_async")(file_path, spritesheet_name, sprites, "[hash_icons]", "[generate_dmi]", "[flatten]")
+/// Creates a single DMI or PNG using 'sprites' as a list of icon states / images.
+/// This function is intended for generating icons with only a few states that have little in common with each other, and only one size.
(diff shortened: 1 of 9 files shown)


AffectedArc07:
Is this iconforge getflaticon?

itsmeow:
no...?

PR #244, opened 2025-11-13
Prevent `decode_base64` and `decode_base32` from panicking if given invalid base64

Absolucy:
title. these shouldn't unwrap, else it panics and all hell breaks loose.

i was gonna rewrite this to use the `data-encoding` crate instead of `base64` and `base32`, and add url-safe + nopad base64 support, but I'm lazy, so this will do for now.


ZeWaka (commented):
And what will this return on failure? This should probably be codedocced

PR #262, opened 2026-05-04
Adds a new procedural generation algorithm that will be used to generate a more cave-system like Lavaland

CabinetOnFire:
This PR adds a layered terrain generation system that allows us to generate nice looking connected caverns.

It essentially performs the following steps to generate terrain:
1.  Generate Binary Space Partitioning with limits on ratio and size to split the area up into nice leaves of space
image

2. Create a room/rectangle of a random scale with random offsets in each of these leaves. These are the pockets that will be used to generate the actual cave pockets.

image

3. Generate corridors between these pockets
image

4. Apply cellular noise to the result of the above. Except instead of normal cellular noise where things are either dead or alive, we add 2 more states "Definitely dead" and "Definitely alive". Neighboring tiles treat these as if they were just dead or alive. but the tiles themselves are immutable. This lets us ensure that the center of the pocket and the corridors cannot be changed, and means the "golden path" is always going to be floor rather than wall.
image

Along side this, I added support to pass in a prefab json. This JSON lets you pass in data regarding structure that already exist in the grid, and should be either treated as definitely dead or definitely alive. This lets the ruins influence the generation to make them look more natural. 

This looks like the following in DM:
`list(
			"cx" = center_turf.x,
			"cy" = center_turf.y,
			"w" = active_ruin.width,
			"h" = active_ruin.height,
			"isEnclosed" = active_ruin.enclosed_for_terrain,
		)`
		
cx = center of ruin x
cx = center of ruin y
w = width of ruin
h = height of ruin
isenclosed = Whether the ruin is definitely alive (not enclosed) or definitely dead (enclosed)

Finaly result in TGstation looks like below
https://streamable.com/vgduo2

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,11 +26,11 @@ thiserror = "2.0"
 flume = { version = "0.12", optional = true }
 chrono = { version = "0.4", optional = true }
 base64 = { version = "0.22", optional = true }
-md-5 = { version = "0.10.6", optional = true }
+md-5 = { version = "0.11.0", optional = true }
 twox-hash = { version = "2.1", optional = true }
 const-random = { version = "0.1.18", optional = true }
-sha-1 = { version = "0.10", optional = true }
-sha2 = { version = "0.10", optional = true }
+sha1 = { version = "0.11.0", optional = true }
+sha2 = { version = "0.11.0", optional = true }
 hex = { version = "0.4", optional = true }
 percent-encoding = { version = "2.3", optional = true }
 url-dep = { version = "2.5", package = "url", optional = true }
@@ -38,7 +38,7 @@ png = { version = "0.18", optional = true }
 image = { version = "0.25", optional = true, default-features = false, features = [
     "png",
 ] }
-gix = { version = "0.81", optional = true, default-features = false, features = [
+gix = { version = "0.83", optional = true, default-features = false, features = [
     "revision",
     "max-performance-safe",
     "sha1", # ideally switch to sha256 someday
@@ -67,6 +67,7 @@ ammonia = { version = "4.1", optional = true }
 fast_poisson = { version = "1.0.2", optional = true, features = [
     "single_precision",
 ] }
+kiddo = { version = "=5.2.4", optional = true }
 symphonia = { version = "0.5.5", optional = true, features = ["all-codecs"] }
 caith = { version = "4.2.4", optional = true }
 uuid = { version = "1.23", optional = true, features = [
@@ -76,7 +77,7 @@ uuid = { version = "1.23", optional = true, features = [
 ] }
 cuid2 = { version = "0.1.4", optional = true }
 rand_chacha = { version = "0.10.0", optional = true }
-hmac = { version = "0.12.1", optional = true }
+hmac = { version = "0.13.0", optional = true }
 base32 = { version = "0.5.1", optional = true }
 indexmap = { version = "2.14", optional = true, features = [
     "serde",
@@ -161,7 +162,7 @@ hash = [
     "hmac",
     "rand",
     "rand_chacha",
-    "sha-1",
+    "sha1",
     "sha2",
     "twox-hash",
     "serde",
@@ -198,7 +199,7 @@ uuid = ["dep:uuid", "cuid2"]
 dice = ["caith"]
 ed25519 = ["base64", "ed25519-dalek", "rand", "zeroize"]
 pathfinder = ["num-integer", "pathfinding", "serde", "serde_json"]
-poissonnoise = ["fast_poisson"]
+poissonnoise = ["fast_poisson", "kiddo"]
 redis_pubsub = ["flume", "redis", "serde", "serde_json"]
 redis_reliablequeue = ["flume", "redis", "serde", "serde_json"]
 unzip = ["zip", "jobs"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,7 +67,6 @@ ammonia = { version = "4.1", optional = true }
 fast_poisson = { version = "1.0.2", optional = true, features = [
     "single_precision",
 ] }
-kiddo = { version = "=5.2.4", optional = true }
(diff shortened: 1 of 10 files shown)


CabinetOnFire:
░░░░░░░░░░░░▄▄
░░░░░░░░░░░█░░█
░░░░░░░░░░░█░░█
░░░░░░░░░░█░░░█
░░░░░░░░░█░░░░█
███████▄▄█░░░░░██████▄
▓▓▓▓▓▓█░░░░░░░░░░░░░░█
▓▓▓▓▓▓█░░░░ Zewaka, ░░░░█
▓▓▓▓▓▓█░░░░Add This░░░░█
▓▓▓▓▓▓█░░░░Please!░░░░░█
▓▓▓▓▓▓█░░░░░░░░░░░░░░█
▓▓▓▓▓▓█████░░░░░░░░░█
██████▀░░░░▀▀██████▀

src/lavaland_generator.rs
```diff
@@ -0,0 +1,636 @@
...
+// ─── Input Structs ─────────────────────────────────────────────────────────────
+
+fn deserialize_byond_bool<'de, D>(deserializer: D) -> std::result::Result<bool, D::Error>
+where
+    D: serde::de::Deserializer<'de>,
+{
+    match u8::deserialize(deserializer)? {
+        0 => Ok(false),
+        _ => Ok(true),
+    }
```
Absolucy: really doesn't matter at all, just looks cleaner imo. but this is purely opinion.
```suggestion
    u8::deserialize(deserializer)?.map(|x| x != 0)
```

src/lavaland_generator.rs
```diff
@@ -0,0 +1,636 @@
...
+    let dy = y1 as f64 - y2 as f64;
+    (dx * dx + dy * dy).sqrt()
+}
+
+fn kruskal_mst(n: usize, edges: &[MSTEdge], loop_percent: usize, rng: &mut impl Rng) -> Vec<MSTEdge> {
+    let mut sorted = edges.to_vec();
+    sorted.sort_by(|a, b| a.dist.partial_cmp(&b.dist).unwrap());
+
+    let mut parent: Vec<usize> = (0..n).collect();
+    let mut result = Vec::new();
```
Absolucy:
```suggestion
    let mut result = Vec::with_capacity(sorted.len());
```

Absolucy (changes requested):
nowhere near a comprehensive review, just two surface-level improvements

src/lavaland_generator.rs
```diff
@@ -0,0 +1,633 @@
...
+struct Room {
+    x: usize,
+    y: usize,
+    w: usize,
+    h: usize,
+    cx: usize,
+    cy: usize,
+}
+
+#[derive(Clone)]
```
ZephyrTFA: This can be Copy instead of Clone

src/cave_system_generator.rs
```diff
@@ -0,0 +1,633 @@
...
+
+    // Step 4: Adjacency edges + Kruskal MST
+    let edges     = build_adjacency_edges(&leaves);
+    let mst_edges = kruskal_mst(leaves.len(), &edges, loop_percent, &mut rng);
+
+
+
+    // Step 5: Apply rooms to grid as DEF_ALIVE (to prevent them from being eaten by noise), skipping prefab-fixed cells
+    for leaf in &leaves {
+        if let Some(ref room) = leaf.room {
```
ZephyrTFA, line 179: This could probably be an iter for each

src/lib.rs
```diff
@@ -66,6 +66,8 @@ pub mod url;
 pub mod uuid;
 #[cfg(feature = "worleynoise")]
 pub mod worleynoise;
+#[cfg(feature = "lavaland_generator")]
```
ZephyrTFA: Alphabetical order

Cargo.toml
```diff
@@ -140,6 +141,7 @@ all = [
     "redis_reliablequeue",
     "unzip",
     "worleynoise",
+    "lavaland_generator",
```
ZephyrTFA: Alphabetical

Cargo.toml
```diff
@@ -200,6 +202,7 @@ redis_pubsub = ["flume", "redis", "serde", "serde_json"]
 redis_reliablequeue = ["flume", "redis", "serde", "serde_json"]
 unzip = ["zip", "jobs"]
 worleynoise = ["rand", "rayon"]
+lavaland_generator = ["rand", "rayon", "serde", "serde_json"]
```
ZephyrTFA: Alphabetical

ZephyrTFA (changes requested):

src/lib.rs
```diff
@@ -36,6 +36,8 @@ pub mod http;
 pub mod iconforge;
 #[cfg(feature = "json")]
 pub mod json;
+#[cfg(feature = "lavaland_generator")]
+pub mod lavaland_generator;
```
ZeWaka: It should probably be called something else not specific

CabinetOnFire: Renamed to cave system generator

CabinetOnFire:
@SmArtKar has threatened to beat me if I don't get this PR merged. :*( they are so violent

ZeWaka:
> @SmArtKar has threatened to beat me if I don't get this PR merged. :*( they are so violent

Tempting to delay so the beatings continue... >:)

SmArtKar:
> > @SmArtKar has threatened to beat me if I don't get this PR merged. :*( they are so violent
> 
> Tempting to delay so the beatings continue... >:)

My wrath shall fall upon you both (this isn't urgent as we can't actually update rust-g past 4.something on /tg/ anyways, so the PR that requires this will be on hold until then)

ZeWaka (approved):

ZeWaka:
clippy warnings

ZeWaka:
sorry forgot about this will look today

Kapu1178:
streamable is a shit website pls embed video directly

CabinetOnFire:
github always shits itself when I try to upload videos.

PR #267, opened 2026-07-26
makes `file_write` atomic

Absolucy:
changes `file_write` to an atomic write - the actual file will not get truncated/changed until the write is successful

basically like. you won't lose the original file if the disk/filesystem has a fucky wucky - we actually ran into this issue on monke, where a lot of prefs were lost due to running out of disk space, and thus prefs json files just got truncated and couldn't be written to.

it works by writing to a tempfile first - then, once the write succeeds, it just renames the tempfile to the target filename. simple as that.

also got rid of that "this is god-awful" thing and replaced it with just an `into_error()`


SpaceManiac:
Is there a reason to ever want the non-atomic behavior? Why two functions?

Absolucy:
backwards compat, mostly. why aren't all file writes atomic on the os?

basically i wanted to avoid messing with pre-existing behavior.

Absolucy:
@SpaceManiac actually technically it is a breaking change

`file_write` returns how many bytes were written

ZeWaka (commented):
Honestly, i'd rather the breaking behavior change than keep a footgun in the API.
I don't see how you can't just keep the bytes in the return value.

ZeWaka (approved):

PR #268, opened 2026-08-02
Adds support for A* pathfinding via DLL using baked turf info

CabinetOnFire:
Sibling PR to https://github.com/tgstation/tgstation/pull/97312

I'm implementing a system on /tg/ to store information about navigation-ability on turfs, and sending that info to rust so we can run a modified A* over it to reduce pathfinding costs. 

Each turf stores (via a bitfield)
- Are we baked or not
- Is this turf simulated (atmos)
- Which directions would this turf let a ground mover pass
- Which directions would this turf let a flying mover pass
- Which directions have a conditional blocker (e.g. door, tables, flaps)

and stores a list of what blockers exist in each direction from a turf, allowing us to call canastarpass on them.

The pathfinding can either be done blocking (instantly gives a path) or async. There's also methods to send the cached data to rust-g which it uses to run the pathfinding cheaply.

This uses ByondAPI (for the conditional blockers), I tried integrating the crate for it and that required adding LLVM for windows. I havn't tested on linux/mac and I'm not sure exactly how it would work on those tbh. I struggled with this a bit so if someone knows if I fucked up lmk. Also let me know if there's concerns with integrating byondapi 😳


CabinetOnFire:
not sure what to do with the 64 bit builds, byondapi doesn't support it but also not sure why it's needed. would love 2 cents on this : )

Iamgoofball:
@CabinetOnFire 64 bit builds are for OpenDream, worth checking in with them on that

ZeWaka (changes requested):
At this time, I'm not comfortable with incorporating anything byondapi into this codebase. 
-maintainer

CabinetOnFire:
All good, I will migrate this to a separate dll! :D

PR #269, opened 2026-08-03
adds configurations  and a README entry for debugging

Phantastic-Swan:
To make it easier for other in the future to debug this library, I've added a few json entries and a README entry about how to debug it.

I've also added entries in the `tasks.json` file to allow for compiling in development mode (no compiler optimizations) and in release mode (with compiler optimizations).

I also added the `ms-vscode.cpptools` VSC extension to the list of recommended extensions to make it easier to access it for debugging.


ZeWaka (commented):
Looks good on first glance, will follow the instructions later to verify this exact flow works

Phantastic-Swan:
tell me if something doesn't, this works for me but I wouldn't doubt that something might be specific to my config

ZeWaka (approved):

PR #271, opened 2026-09-16
Adds an option to select the timezone in which logs are written

Phantastic-Swan:
Adds a new argument to the `log_write` function allowing for selection of timezones in which the log will be written.
image

The default timezone for passing an empty string/incorrect value is still UTC, and the servers local timezone can be selected by passing "Local" as the argument. Timezone input should be given as the [IANA timezone name](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones).


ZephyrTFA:
logs were intended to be logged at UTC time regardless of server time.

AffectedArc07:
Logs are normally in UTC to avoid DST hell and timezone confusion for when admins have to piece them together. Hard no from me.

ZeWaka (changes requested):
I would be open to a new optional arg that allows a format string, but as is, this is non-tenable due to concerns listed above by other users.

Phantastic-Swan:
alright, given that this became a bit of a bigger change and this is my first time writing something like this in rust, other than just making sure both the functions work and display logging, is there anything else I should test to make sure I haven't fucked something up?

src/log.rs
```diff
@@ -14,35 +15,51 @@ thread_local! {
...
+            } else {
+                // write first line, timestamped
+                let mut iter = data.split('\n');
+                if let Some(line) = iter.next() {
+                    if utc_time {
+                        writeln!(file, "[{}] {}", Utc::now().format("%F %T%.3f"), line)?;
+                    } else {
+                        writeln!(file, "[{}] {}", Local::now().format("%F %T%.3f"), line)?;
+                    }
+                }
```
AffectedArc07, line 51:
```suggestion
                let tz = if utc_time { Utc::now() } else { Local::now() };
                if let Some(line) = iter.next() {
                    writeln!(file, "[{}] {}", tz.format("%F %T%.3f"), line)?;
                }
```

Disclaimer, I initially thought rust had ternaries then realised it didn't when writing that. Please do test locally before accepting the change, or just write it off. Either or.

Phantastic-Swan: the ternary operator isn't the problem, `Utc::now()` and `Local::now()` not being the same type is. However, this would work:
```rust
let tz = if utc_time { Utc::now().format("%F %T%.3f") } else { Local::now().format("%F %T%.3f") };
if let Some(line) = iter.next() {
     writeln!(file, "[{}] {}", tz, line)?;
}
```
but with how long that first line is I'm not necessarily certain it is better. Especially when the editor hints at the type AND it should be named timestamp for better clarity.

Personally, if anything I'd do something like this, and I would imagine the compiler should optimise it away into the same thing as above.
```rust
let utc_timestamp = Utc::now().format("%F %T%.3f");
let local_timestamp = Local::now().format("%F %T%.3f");
let timestamp = if utc_time { utc_timestamp } else { local_timestamp };
if let Some(line) = iter.next() {
    writeln!(file, "[{}] {}", timestamp, line)?;
}
```
lmk which one of those you would prefer me to go with, if any.

AffectedArc07: image

Both are `chrono:offset` types. You'd need to cast them down to another type, but again, not my call to make.

Phantastic-Swan: not according to my rust
image
image
image
image

AffectedArc07: Each one is its own type yes but the share the parent type of `chrono::offset`.

dmsrc/log.dm
```diff
@@ -1,2 +1,5 @@
+/// rust_g function to write to a log file. Timestamps in the UTC timezone
 #define rustg_log_write(fname, text, format) RUSTG_CALL(RUST_G, "log_write")(fname, text, format)
+/// rust_g function to write to a log file. Timestamps in the local timezone (same as the server)
+#define rustg_log_write_local_time(fname, text, format) RUSTG_CALL(RUST_G, "log_write_local_time")(fname, text, format)
```
ZeWaka, line 4: This isn't what I really wanted, I said a time format specifier argument

Phantastic-Swan: ok, what exactly should the format be? timezone shortcut, offset from UTC? Also imo we should still have a separate function for local time just so that if anyone wants to use it they don't have to check which timezone the server is in

ZeWaka: probably the iana time zone, e.g. `America/Vancouver`. you could just pass "local" or "server" for servertime

dmsrc/log.dm
```diff
@@ -1,2 +1,13 @@
-#define rustg_log_write(fname, text, format) RUSTG_CALL(RUST_G, "log_write")(fname, text, format)
+/**
+ * rust_g function to write to a log file.
+ * 
+ * Arguments:
+ * * fname - path to the log file from the root of the directory
+ * * text - the log message
+ * * timezone - the timezone in which the time should be given. Defaults to UTC for an 
+ * empty string/incorrect value. Timezone must be given as a TZ identifier (https://en.wikipedia.org/wiki/List_of_tz_database_time_zones).
+ * You can also input "Local" to use the servers local time
```
ZeWaka, line 9: I would really prefer timezone be after format so this isn't an annoying breaking change

also unclear if `Local` needs to be capitalized in that description, use backticks if it's intentional

Phantastic-Swan: by backticks you mean `? the things that do the code line/block things?

Phantastic-Swan: also, as far as I know, you cannot input 3 arguments for a macro that needs 4. I could revert the change to this macro so it remains untouched, and have it input `UTC` by default, and then create another macro to allow for timezone input

ZeWaka: > also, as far as I know, you cannot input 3 arguments for a macro that needs 4.

yes. but as is, people need to shuffle their arguments instead of just adding the timezone

you can have a varadic macro length using `..rest` - pretty sure we use it elsewhere in rustg

> by backticks you mean `? the things that do the code line/block things?

yes https://en.wikipedia.org/wiki/Backtick

